PR #7, opened 2019-12-27
Add CI setting

andy840119:
Copied from osu.game


andy840119:
At least base CI works.
Now can check testcase in pr

PR #8, opened 2019-12-28
Update

andy840119:
Update


andy840119:
Add
```yml
branches:
  only:
    - master
```

to prevent build twice.

Lazy to create another pr.

See [here](https://github.com/shibayan/appservice-acmebot/pull/86/files)

PR #11, opened 2019-12-28
Upgrade VS version to 16.4

Squiddu:
An upgrade to at least 16.4 will improve security and performance. [Here are the patch notes for the base 16.4 version.](https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes#--visual-studio-2019-version-164)


andy840119:
Thanks for contribution 👍

PR #15, opened 2020-01-29
Update to latest progress

andy840119:
Update to latest progress


andy840119:
- Add microphone support
- Change header
- Update nuget package

andy840119:
Make a merge now ?
Because this bug cannot be fixed recently and easy to be tracked.

PR #39, opened 2020-03-18
Update latest ppy.game

andy840119:
Update latest ppy.game


andy840119:
😠

andy840119:
Add `ILegacyRuleset` now for preventing `KaraokeRulesetConfigManager` is null if no legacy id
Will be removed.

Issue #56, opened 2020-04-09
Fix suggestion rider hint


andy840119:
![image](https://user-images.githubusercontent.com/9100368/78902336-a0e0a400-7ab4-11ea-8858-4df239a30ade.png)
Fix this part of setting.
Should be fixed in change config.

andy840119:
Implemented in #62

PR #62, opened 2020-04-11
Fix suggestion rider hint

andy840119:
Implementation of #56


andy840119:
Second CI test!

andy840119:
Write some comment here to mark how this doing.

Issue #59, opened 2020-04-10
Toogle hit in karaoke note


andy840119:
![image](https://user-images.githubusercontent.com/9100368/79000152-1321b900-7b87-11ea-91ba-155bcab7b54b.png)

See how this working.

andy840119:
Implementing

PR #63, opened 2020-04-11
Toogle hit in karaoke note

andy840119:
Implementation of #59 
.
TODO : 
- [x] Add testcase
- [x] Add blink layer in skin component
- [x] blink effect should looks good in testcase.

### DIFF
--- a/osu.Game.Rulesets.Karaoke.Tests/Skinning/KaraokeHitObjectTestScene.cs
+++ b/osu.Game.Rulesets.Karaoke.Tests/Skinning/KaraokeHitObjectTestScene.cs
@@ -41,7 +41,11 @@ namespace osu.Game.Rulesets.Karaoke.Tests.Skinning
                             Clock = new FramedClock(new StopwatchClock()),
                         }.With(c =>
                         {
-                            c.Add(CreateHitObject().With(h => h.AccentColour.Value = Color4.Orange));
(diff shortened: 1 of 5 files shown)


andy840119:
Bad codefactor 🤔

andy840119:
How to implement sparkle effect in legacy skin component?
1. Add new layer
2. Display list of image, should be rename at the end like `-0`, `-1`. See `LegacyBodyPiece` and [osu.Game.Resources/Skins/Legacy](https://github.com/ppy/osu-resources/tree/master/osu.Game.Resources/Skins/Legacy) for more detail.

andy840119:
If skin resource is not end of `-0`  `-1`, `skin.GetAnimation(imageName, true, true)` will auto convert into `Sprite`.
Can create sparkle layer or display animate effect, but when should use whicn?
Or just use both?
.
Use both seems faster.
If animate exist, then display.
If sparkle layer exist, then sparkle with `default/target` colour
If (!animate && !sparkle layer), then ignore or use default layer?

andy840119:
In the last comment, how to define `animate is not exist`?

andy840119:
Don't care about animate is exist.
Just `sparkleLayer ?? defaultLayer`.
If you don't want sparkle effect, add `DisableSparkle` config.

PR #64, opened 2020-04-12
Update nuget

andy840119:
- [x] Wait until #63 is merged


andy840119:
TODO : it will cause crash because karaoke ruleset cannot accept another ruleset's beatmap

PR #70, opened 2020-04-18
20200418 implement lyric editor playfield

andy840119:
Implement playfield

PR #71, opened 2020-04-19
20200418 lyric editor screen

andy840119:
In compare of #70 using playfield to deal with lyric editor, maybe it should be better to use new screen to deal with it.


andy840119:
Merge it now to let this project has little progress

Issue #73, opened 2020-04-25
Make Note into lyric's nested object


andy840119:
Inspire from #69 
For better management in the lyric maker, note should be lyric's nested object.
.
How to : 
1. Initial created note in encoder should be stored into `NestedHitObjects`.
2. ApplyDefaultsToSelfshould save all `nestedHitObjects` because `ApplyDefaults` will clear all nasted object, then restore it with `CreateNestedHitObjects`(or have better way?)
3. `CreateNestedHitObjects` should apply note's real time by timetag's time
4. Think about how to create `drawable note` and put it's into drawable
5. Judgement should work fine.
6. This branch can be merged into master.
.
Notice :
See how `Slider` do in osu ruleset.

andy840119:
Don't need to think changing TimeTags or Note's number

andy840119:
This issue can be closed because :
- `Note` and `Lyric` no need strong related.
- Not need to do this because can use like `lyricManager` to delete notes if lyric has been deleted.
- `Note` has many property that cannot be saved into `Lyric`.

PR #74, opened 2020-04-26
Make Note into lyric's nested object

andy840119:
Implementation of #73


andy840119:
Note : 
`AddNestedHitObject` is called after `LoadComplete`, and `LoadComplete` only called when hitobject display.

andy840119:
Maybe to this in another day because `nestedHitObjects` will be force added in  `private void apply(HitObject hitObject)` if `hitObject.NestedHitObjects ` is not empty.
And will some trouble if `NestedHitObjects` has value.

andy840119:
Do not do that because it might let note object become larger.
And json formatter enable to save note's parent lyric by reference.

PR #82, opened 2020-05-08
Little clean-up

andy840119:
Little clean-up code


andy840119:
\Wow code factor is working/

Issue #85, opened 2020-05-13
Hide HUDOverlay is beatmap is not scorable


andy840119:
1. add an extension to check this beatmap can be scored or not (check by has any note or not).
2. Hide the HUDOverlay if `1.` matched or Mod's `ScoreMultiplier == 0`(need to check).

andy840119:
Also need to create `KaroakeHUDOverlay` and it's own interface(`IApplicableToKaroakeHUD`) and method(`ApplyToKaraokeHUD(KaroakeHUDOverlay overlay)`)

andy840119:
maybe put `KaroakeHUDOverlay ` into `drawableRuleset.Overlays`?

andy840119:
Separate into two PRs.
First PR :
1. create `IApplicableToKaraokeHUD` then let mods can access `KaraokeHUDOverlay : Container`.
2. give `ControlOverlay` a better name , public readonly in `KaraokeHUDOverlay`.
3. create file `DrawableKaraokeRulesetHUD.cs` to deal with HUD logic ( partial of `DrawableKaraokeRuleset`)
4. remember to adjust test case.

andy840119:
Implemented

PR #86, opened 2020-05-16
Create karaoke HUD

andy840119:
Make karaoke ruleset has it's own control  HUD layer.
Implement part of #85


andy840119:
Seems ok.
Make a merge.

PR #87, opened 2020-05-16
Add beatmap extension to deal with duplicated logic

andy840119:
1. add karkaoke beatmap extension
2. rename translate dicrionary into property dictionary
3. move some duplicated code into extension.


andy840119:
hmm...
will deal with it future.

PR #96, opened 2020-05-30
Improve saiten performance

andy840119:
- Update nuget package and fix trigger event


andy840119:
It's good enough to merge it now 😮 
But i'm not sure it's better to wait until https://github.com/osu-Karaoke/osu-framework-microphone/pull/29 has merged.

andy840119:
Seems ok to merge.

PR #98, opened 2020-05-31
Implement singer autoplay mod

andy840119:
Implement singer autoplay mod


andy840119:
TODO : give empty replay if song is not saitenable(scorable)

andy840119:
Test case is added and seems great.
But i still have no idea why song's result is bad 😢

andy840119:
Consider auto generator has test case now.
Merge it now and leave another pr to fix that.

PR #105, opened 2020-06-13
Implement KaraokeReplayRecorder

andy840119:
Create karaoke recorder class to convert list saiten action into replay frame.


andy840119:
I'm not sure it's better to convert from legacy frame or not because karaoke beatmap always not showing result, which not actually create replay.

andy840119:
Legacy replay has been implement.
But replay is created in result page, not in gameplay.

andy840119:
Waiting for CI

Issue #104, opened 2020-06-07
Together saiten conversion logic.


andy840119:
簡單來說
Beatmap有一套紀錄，紀錄原始聲音到可以被踩點需要多少位移和縮放
玩家也可以設定自己的位移和縮放，不然有些音太高或是太低會唱不上去
.
而在replay record時也會記錄下來當前的位移縮放
.
所以
平常遊玩時 :
玩家聲音 -> [Beatmpa紀錄換算] -> [玩家記錄換算] -> 顯示在採點UI上的數據
回播時
玩家聲音 -> [Beatmpa紀錄換算]  -> [Replay季錄換算] -> 顯示在採點UI上的數據
大概是這樣
.
有點難用英文表示...

andy840119:
在圖譜中，不管裡面的歌手是男生或是女生
都會被轉換成中性音階，避免Tone整體偏高或是偏低
.
然後遊戲設定會根據玩家是男生還是女生給出對應的設定
至於要不要微調就看玩家決定
.
然後設定會被保存在replay中

andy840119:
`KaraokeSaitenAction`'s Scale should change name to pitch and should convert to tone until `calculator.YPositionAt(action)`

andy840119:
But how to ruleset know is in replay or playing?

andy840119:
This change might modify `AutoGeneratorBySinger`'s test case.
Make a json file to compare result instead of compare it in code(compare json string).
.
Also, rename `demo.mp3` to `auto-generator-test-case.mp3`, and add `auto-generator-test-case.json`.

andy840119:
Oops.
Singer's pitch to tone should be converted in `KaraokeAutoGeneratorBySinger`.
Because:
1. Much simple.
2. Didn't affect `KaraokeAutoGenerator`'s logic. Also `KaraokeAutoGenerator` don't need conversion.
3.

andy840119:
TODO : 
1. add a extension to deal with convert pitch to scale.
2. add a extension(to beatmap) to convert raw sacle to beatmap's scale.
3. add a fake position and zoom adjustment value in beatmap.
Note : 
need to write notes to list all the conversion.

PR #107, opened 2020-06-13
Together saiten conversion logic.

andy840119:
Implementation of #104 
Also : 
- Use json to record target pitch record in testcase.
- And collect pitch to scale logic.


andy840119:
Maybe pass two or three demo pitch song(with BGM)?
But it's not important issue so can be ignore now.

PR #116, opened 2020-06-20
Slave/fix change log

andy840119:
- Change organization name.
- Fix any relative error.


andy840119:
Found switch tab cause crash.
皮皮的鍋，懶得修了

PR #117, opened 2020-06-21
Parse tag in change log

andy840119:
1. Separate `ChangeLogMarkdownContainer`
2. Parse special pull-request(e.g : `#123`) and mention tag (e.g : `@andy840119`).
3. Add a simple test case.


andy840119:
TODO : 
1. Should add list of white-list to let user to pares special string they want.

PR #125, opened 2020-06-27
Support SynthV project file.

andy840119:
Support SynthV project file.


andy840119:
Need to have nuget package for parsing this kind of format.

Issue #130, opened 2020-07-01
Add github action to auto-deploy to release.


andy840119:
See how [this project](https://github.com/LumpBloom7/sentakki/releases) do.

andy840119:
Implemented in #132 
Wait until make a success release build.

andy840119:
it works!

PR #131, opened 2020-07-04
Create dotnet-core.yml for github CI action.

andy840119:
Part of implementation of #130 .
Not sure is needed because appveyor is handling this job.


andy840119:
Maybe waiting until appveyor cannot be used?

PR #132, opened 2020-07-04
Add release setting

andy840119:
Mark that this code is copied from the https://github.com/LumpBloom7/sentakki/commits/master/.github/workflows/release.yml


andy840119:
Maybe merge and see what's happen 🤔 .

PR #135, opened 2020-07-05
Fix lyric's start time issue.

andy840119:
1. Create `IPatternGenerator` to batch calculate like calculate karaoke text's layout and start time.
2. Fix lyric's timing issue, need to adjust duration if want to move start time.


andy840119:
Might one tests failed.
Ignored.

PR #138, opened 2020-07-07
Update package

andy840119:
Update ppy.game and mocriphone package.
Mocriphone has update framework and tested.
TODO : 
Need to fix some test error.
Still no idea.

### DIFF
--- a/osu.Game.Rulesets.Karaoke/KaraokeInputManager.cs
+++ b/osu.Game.Rulesets.Karaoke/KaraokeInputManager.cs
@@ -58,7 +58,7 @@ namespace osu.Game.Rulesets.Karaoke
         }
 
         protected override InputState CreateInitialState()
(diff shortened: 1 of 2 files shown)


andy840119:
Use
```Csharp
public class KaraokeRulesetInputManagerInputState<T> : RulesetInputManagerInputState<T>, IMicrophoneInputState
        where T : struct
    {
        public MicrophoneState Microphone { get; }

        public KaraokeRulesetInputManagerInputState(MicrophoneInputState state)
            : base(state)
        {
            Microphone = state.Microphone;
        }
    }
```
Instead in `KaraokeInputManager`.
Also need to update karaoke package.

andy840119:
Microphone cannot open now
Need to find the reason.

andy840119:
Fixed.

Issue #148, opened 2020-07-26
Fix assembly not found issue in release


andy840119:
![image](https://user-images.githubusercontent.com/9100368/88476878-bcde5780-cf76-11ea-86c6-33c2226ee191.png)
It's caused by some relative nuget's dll is not imported.
Maybe see how [gamebosu](https://github.com/Game4all/gamebosu/blob/master/osu.Game.Rulesets.Gamebosu/osu.Game.Rulesets.Gamebosu.csproj) do?

andy840119:
![image](https://user-images.githubusercontent.com/9100368/88478295-16984f00-cf82-11ea-8d36-e191be6cfbb1.png)

Need to combine those file and make a check if will crash in debug/release test case.
Also include debug point is working or not?

File need to be packed: 
- LyricMaker
- NicoKaraParser
- NWaves
- Octokit
- osu.Framework.Microphone
- osu.Game.Rulesets.Karaoke
- osu.KaraokeFramework

andy840119:
Hmm...
The better way might to merge those dlls to single one in github action.

andy840119:
Also need to try github action works ok or not.
https://github.com/nektos/act

andy840119:
Implemented and tested.

PR #151, opened 2020-07-27
Fix assembly not found issue in release

andy840119:
Implementation of #148 .
Will look at what's wrong.
Should test on mac also.


andy840119:
Copy from : https://github.com/Game4all/gamebosu/blob/master/osu.Game.Rulesets.Gamebosu/osu.Game.Rulesets.Gamebosu.csproj

andy840119:
image
On mac got this alert, but still works 🤔 
Need to place it in mac version's lazer.

andy840119:
image
Mac version works.
But lost lcon.
Maybe need to clean record in my PC then try again.

andy840119:
Ignore icon does not display issue.

PR #154, opened 2020-08-01
Implement microphone status preview.

andy840119:
In order to let player understand microphone is set-up or not.
It's need to make a visual part in game play to let user understand microphone status and how to open microphone.


andy840119:
TODO : 
1. might get microphone status from` KaraokeSessionStatics` in `KaraokeInputManager`
2. `NotePlayfield` might be lazy-create until `BDL`
3. Add several type in `SaitenStatusMode` if microphone is not open due to some issue like: 
    - `NotSaitening`: Beatmap is not scoring.
    - `Auto play`: Auto play.
    - `Edit`: Editor mode, saiten mark should be hidden(or not loaded) in `NotePlayfield`

andy840119:
Need to make a tutorial about how to get microphone permission.
But it should be wait until all tutorial has been written.

PR #158, opened 2020-09-03
Update osu and fix things that broke

LumpBloom7:
Super knee jerk fix, so that a update can be done ASAP.


andy840119:
Seems it can compile success.
And change log test ok.
Make a fast fix not without testing gameplay.

Issue #168, opened 2020-09-13
Cannot open setting dialog if not in practice mod.


andy840119:
Cannot open setting dialog if not in practice mod.

andy840119:
Refactor overlay in this pr.

andy840119:
Seems only `ModAutoplay` can trigger button now.

andy840119:
Implemented.

PR #171, opened 2020-09-13
Fix setting dialog cannot be click if not in auto-play mod.

andy840119:
Fix issue found in #168 

TODO : 
- [x] Refactor the code.
- [x] Search the reason.


andy840119:
Both of mouse/keyboard can trigger `ToggleVisibility`.
But overlay not display 🤔

andy840119:
Have no idea now 🤔 
maybe it's caused by overlay?

Issue #177, opened 2020-09-23
Create scoring result graph


andy840119:
Let user can see which note singing wrong.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/94027794-35865680-fdf6-11ea-8f0c-a37862f7259f.png)
Also add this effect(RankingChartLineGraph).
Hover to see:
- lyric
- note's lyirc text
- Score

andy840119:
In this pr, can pass line graph because cannot get replay.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/94028924-45526a80-fdf7-11ea-9b34-101ceb845a54.png)
View in 1920*1080

andy840119:
Need to re-design it in 16:9 or 4:3 ratio.

PR #179, opened 2020-09-25
Ranking result component

andy840119:
Implementation part of #177


andy840119:
Notice : 
image
image
Also should notice that statistic will in single line if can be.

andy840119:
image
And this is the arrange.

andy840119:
image
Each statistic should throw into different `StatisticRow`

andy840119:
Can make a merge.
But still need to adjust in the future.

andy840119:
image
With two saiten result.

PR #180, opened 2020-09-25
update nuget package.

andy840119:
update nuget package to latest version.

.
Notice : 
Only test case's editor can access now.

.
TODO : 
- [x] Fix ruleset button cannot apply setting now 😢


andy840119:
image
Now looks like that 🤔 
I guess some day will roll-back because it should be UI layer, not toggles.

PR #182, opened 2020-09-26
Re-write tone's calculation test case.

andy840119:
Use test case with value instead of writing multi logic with single test case.


andy840119:
Not really sure will got more star if re-write this test case 🤔

Issue #186, opened 2020-10-01
Beatmap info in result page.


andy840119:
Implement part of #177 
See `BeatmapInfoWedge` and make a similar one.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/94757490-542bb500-03d5-11eb-958c-e6a1dc52a94a.png)
1. Add this in left side
![image](https://user-images.githubusercontent.com/9100368/94757503-5f7ee080-03d5-11eb-9473-212524af4eef.png)
2. Add this in right side
![image](https://user-images.githubusercontent.com/9100368/94757597-a53ba900-03d5-11eb-95d4-5356695e3435.png)
3. If can, also add `Add to collection` button.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/94757553-8b01cb00-03d5-11eb-9188-197f77c2caf5.png)
Also click left side can show song info?

andy840119:
Beatmap info has been implemented.
Remain beatmap description.

andy840119:
Remain singer tooltip.
This issue can be closed.

PR #187, opened 2020-10-01
Implement beatmap info in result page.

andy840119:
image
Saitenable beatmap.

image
Not saitenable beatmap.

Implementation of #186


andy840119:
image

Also click beatmap info can show overlay.

PR #188, opened 2020-10-01
Implement beatmap metadata statistic.

andy840119:
Implement last part of #186 .


andy840119:
image
This is how it looks now.

Issue #141, opened 2020-07-14
Introduce : sub-singer


andy840119:
This is used to present singing status, like singing with echo, background or yelling.

Should be like : 
- Singer 1
  - background
  - yelling
- Singer 2
  - background
  - sad voice

In `editor`, mapper can select `singer` or `status`, e.g : can select `Singer 1` or `background`(Belongs to singer 1).

And alert user not to create singing state like `normal` or `main`.

`Singer` object already means that.

.

Notice : 

This status not means you can mix multi singer with single status.

If you wants multi singer to sing single lyric.
you should select like : 
`Singer 1` + `Singer 2` or `Singer 1` + `background(belongs to singer 2)` or  `background(Singer 1)` + `background(Singer2)`

But mapper cannot select like
`Singer 1` + `background(Singer 1)` or `background(Singer 1)` + yelling(Singer 1)

.

Will be implemented if i had enough time.

andy840119:
Also note what property` singer` and `sub-singer` will have : 

Singer : 
- ID(Hide)
- Display order(hide)
- Name(with any language)
- English name
- Romaji name
- Avatar sticker
- Some basic info like age or weight(optional)?

Sub-singer : 
- ID (start from 10)
- Display order(hide)
- Parent singer's id
- Status(Text)
- Override sticker(Not sure)
- Mix sticker(Will be the cover in parent singer's sticker)

.

And notice that max singer is 10 now, sub-singer is not limited right now.

andy840119:
Also should design a page to show which singer sing which lyric?
Or place it into lyric maker?

andy840119:
Also, should create auto-fill to fill the singer info created in database.

Need to create list of singer in database
- Singer's name and info.
- Some common status text.

andy840119:
Will wait until most of design in editor has been implemented.

andy840119:
Also should notice some mapper set singer as animate or virtual character.
But others might set as real singer.
.
Need to deal with that also.

andy840119:
`SubSinger` Will be the parent of `Singer`
Like 
```
List<SubSinger> SubSingers {get;set;}
```
in `Singer.cs`?

andy840119:
Model has been implemented.
Remain singer editor.

andy840119:
Also should limit max singer number(include sub singer) to 30.
Singer : 10
SubSinger : 30
Because `int.max` is 2,147,483,647, which very close to `2^30(1,073,741,824)`

[...]

PR #194, opened 2020-10-04
Implement remove singer in metadata.

andy840119:
Implement part of #141 
Add basic remove singer and test.


andy840119:
And `SingerMetadata`'s some function should be move and works like `IEditorChangeHandler` in editor eventually.

PR #195, opened 2020-10-04
Implement singer tooltip.

andy840119:
Implement remain part of #188


andy840119:
image
Seems perfect now.

Issue #197, opened 2020-10-09
Create layout tooltip


andy840119:
See how lyric is arrangement by tooltip.
Will in : 
- Lyric Editor's move mode.
- Batch lyric layout editor.

andy840119:
Should add test case.
Like what `SingerToolTip` do.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/95544436-0bf43e80-0a36-11eb-955c-cf7c9769d8b2.png)
This is 下-1 layout
![image](https://user-images.githubusercontent.com/9100368/95546030-7d35f080-0a3a-11eb-870e-a91cd317c3c4.png)
This is 下-2 layout

andy840119:
Implemented.

PR #198, opened 2020-10-09
Implement layout tooltip

andy840119:
Implementation of #197 .


andy840119:
image
Now what's test case looks like 👌

andy840119:
Oop.
Code factor not happy again 😢

Issue #217, opened 2020-10-18
Improve layout editor


andy840119:
![image](https://user-images.githubusercontent.com/9100368/96357141-b0f7d100-1132-11eb-9ffe-e6095af410e8.png)

This is how `LyricLineLayout` loooks like.
Should be moved into `Edit/Layout` namespace.
.
Also should made a new design.
.
This issue can be separated into several PRs : 
- [x] Implement initial screen.
- [x] Implement left side's config.
- [x] Implement right side's preview.
- [ ] Adjust right side's lyric preview size.
- [ ] Implement manage list of layout, like how translate editor do.
- [ ] Make all the feature works.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/96357178-38454480-1133-11eb-95b2-a533d238a209.png)
Left side should made style like this (see `LabelledSlideBar` and `LabelledTextBox`).

PR #218, opened 2020-10-18
Implement layout editor(left side).

andy840119:
Implement part of #217 .


andy840119:
image
Before

image
Now how it looks like 👍

andy840119:
`CodeFactor`'s suggestion can be removed because legacy test case will be removed until #217 has fully implemented.

PR #219, opened 2020-10-18
Implement layout editor(right side)

andy840119:
Implement part of #217 .

What's should be done in this PR : 
- [x] Create preview control.
- [x] Display preview screen(should add border also, and should display ratio digit).
- [x] Display preview lyric.
- [x] Implement adjust ratio.
- [x] Implement change preview lyric layout
- [ ] Need to adjust ratio display (can be implemented in next pr).
- [x] Remove old test case.


andy840119:
image
Seems it's impossible to place preview button in the right side(due to UI seems weird).
Maybe move to left-side part and make a section to place preview component?

andy840119:
image
Looks a little bit reasonable.

- [x] Maybe should adjust `LabelledDropdown`'s colour? -> No way because color is fixed in base class.

andy840119:
image
Manager has been implemented.
Only remain binding property 👍 .

andy840119:
Also should notice that preview's lyric 's size should be adjusted.
Need to think about how to implement 🤔 
Or should copy playfield's rule?

andy840119:
image
Now it looks like.

There are two options : 
- Should trigger skin source changed(?) if skin has been modified.
  - Logical.
  - But more resource consuming.
- Override `DrawableLyricLine`'s layout
  - Reduce maintainence
- Use KaraokeSpriteText instead?
  - Translate will disappear.

Also: 
- [ ] Need to adjust ratio display.

andy840119:
Also should modified control's default value 🤔 .
- [x] Maybe separate `CurrentLayout` into `LoadedLayout`(for initial control purpose) and `EditLayout` (for preview purpose).?

andy840119:
It's ok enough to make a merge now.

PR #220, opened 2020-10-22
Grammar and spelling changes

zoogies:
Your English is very good but I thought I would help make it clearer :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![dev chat](https://discordapp.com/api/guilds/299006062323826688/widget.png?style=shield)](https://discord.gg/ga2xZXk)
 
 
-The source code of the `karaoke!` ruleset, for [osu!lazer](https://github.com/ppy/osu).
+The source code of the `karaoke!` ruleset, based off of [osu!lazer](https://github.com/ppy/osu).
 
 ## Status
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Using this repo to create a `PIRATED` karaoke ruleset is absolutely forbidden.
 
 ## Thanks to
 
-- [osu!](https://github.com/ppy/osu) as it's [framework](https://github.com/ppy/osu-framework) for karaoke!
+- [osu!](https://github.com/ppy/osu) and it's [framework](https://github.com/ppy/osu-framework) for karaoke!
 
 - [RhythmKaTTE](http://juna-idler.blogspot.com/2016/05/rhythmkatte-version-01.html) and [RhythmicaLyrics](http://suwa.pupu.jp/RhythmicaLyrics.html), an open-source software used to create lyrics with time tags.
 Parts of the lyrics editor in this ruleset were inspired by them.
--- a/README.md
+++ b/README.md
@@ -36,10 +36,10 @@ Using this repo to create a `PIRATED` karaoke ruleset is absolutely forbidden.
 
 - [osu!](https://github.com/ppy/osu) as it's [framework](https://github.com/ppy/osu-framework) for karaoke!
 
-- [RhythmKaTTE](http://juna-idler.blogspot.com/2016/05/rhythmkatte-version-01.html) and [RhythmicaLyrics](http://suwa.pupu.jp/RhythmicaLyrics.html), an open-source software used to create lyrics with a time tag.
-Some of the lyrics editor in this ruleset is inspired by them.
+- [RhythmKaTTE](http://juna-idler.blogspot.com/2016/05/rhythmkatte-version-01.html) and [RhythmicaLyrics](http://suwa.pupu.jp/RhythmicaLyrics.html), an open-source software used to create lyrics with time tags.
+Parts of the lyrics editor in this ruleset were inspired by them.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,43 +7,40 @@
 [![dev chat](https://discordapp.com/api/guilds/299006062323826688/widget.png?style=shield)](https://discord.gg/ga2xZXk)
 
 
-The source code of `karaoke!` ruleset, running on [osu!lazer](https://github.com/ppy/osu).
```
andy840119, line 10: This repo is based on lazer(and osu-framework), but might only run on osu!

README.md
```diff
@@ -7,43 +7,40 @@
...
-1. If you plan on using this repo for commercial purposes, please contact me(andy840119@gmail.com) to get permission first.
-2. Use this repo to create/use `PIRATED` karaoke beatmap is absolutely forbidden.
+This repo is covered under the [GPL V3](LICENSE) license.
+If you plan on using this repo for commercial purposes, please contact me at (andy840119@gmail.com) to get permission first.
+Using this repo to create a `PIRATED` karaoke ruleset is absolutely forbidden.
 
 ## Thanks to
 
-- [osu!](https://github.com/ppy/osu) and it's [framework](https://github.com/ppy/osu-framework), karaoke! is modified from them.
+- [osu!](https://github.com/ppy/osu) as it's [framework](https://github.com/ppy/osu-framework) for karaoke!
```
andy840119: osu! is based on osu-framework, i'm not sure can use `as`?

README.md
```diff
@@ -7,43 +7,40 @@
...
 
-- [RhythmKaTTE](http://juna-idler.blogspot.com/2016/05/rhythmkatte-version-01.html) and [RhythmicaLyrics](http://suwa.pupu.jp/RhythmicaLyrics.html), an open-source software to create lyric with time tag.
-Some of lyric editor in this ruleset is inspired from them.
+- [RhythmKaTTE](http://juna-idler.blogspot.com/2016/05/rhythmkatte-version-01.html) and [RhythmicaLyrics](http://suwa.pupu.jp/RhythmicaLyrics.html), an open-source software used to create lyrics with a time tag.
```
andy840119: One karaoke song has many timetags.

andy840119:
Also thanks for your contribution and star/

README.md
andy840119: but osu-framework is created because osu!lazer's need 🤔

README.md
zoogies: What do you mean by might? Wouldn’t it only run on osu lazer?

README.md
zoogies: I think that it’s ok to say this because karaoke is made on lazer which is made on osu framework

README.md
zoogies: That is true, but because lyrics is plural here, it implies that each lyric has its own time tag. You are right though I will change to make it clearer

zoogies: Fixed

README.md
andy840119: Oops, might `not` only run on osu!lazer, my bad.
Lazer is under MIT license, so anyone can use the code as they like.

andy840119: So if anyone made their game client base on [osu!lazer's package](https://www.nuget.org/packages/ppy.osu.Game/), karaoke ruleset can run on it.

README.md
andy840119: Hmm...
I'm poor on grammar, but if this means that it's ok.

README.md
andy840119: Maybe it's better to say it's `based on`.

zoogies: Ahh, makes more sense. I will make it clearer

zoogies: Fixed

README.md
zoogies: After realizing that people might make karaoke run on another project based off the osu framework, you were right to put “and” here so I changed it back

README.md
andy840119: `based off of`?

LumpBloom7: Late to the discussion, but "running on" is more correct than "based on".

andy840119: It's ok for me to any of them, as i'm not native speaker🤔.

zoogies: Yeah that’s what I figured initially since it was only running on lazer at the moment, but Andy phrased it so that I thought it might eventually run on another project made from the osu framework so in that case I figured we could say based off of as it fits and wont need to be changed in the future if that ever occurs

zoogies: It is purely preference though, I think running on fits better but based off of/on still works

andy840119: Hmm...
I'm just doesn't know `based off of` means.
If this word (`The source code of the karaoke! ruleset, based off of osu!lazer`) has no grammar issue, then it's ok.

zoogies: Just guessing based off your username but if you know Chinese it is the same as “基于”, it is a different way to say “based on”. I will change it to be “running on” because I agree with bloom it makes more sense

andy840119: Did you know Chinese?
Cool

Issue #223, opened 2020-10-24
Improve style editor


andy840119:
![image](https://user-images.githubusercontent.com/9100368/97069498-cc1f8080-160b-11eb-81be-33b66f6452ad.png)
This is now style looks like.
Should be moved into `Edit/Style` namespace.
.
Also should made a new design.
Should have style like exist translator/layout editor
.
This issue can be separated into several PRs :
- [x] Implement initial screen.
- [x] Implement left side config.
- [ ] Implement right side preview.
- [ ] Implement edit note style.
- [ ] Implement manage list of style.
  - This about how to list all of them.
  - Also should mark this style have how many lyric use(?).
- [ ] Make all the feature wotks.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/96531928-bfc2bd00-12c5-11eb-8402-021caff76886.png)
Might have looks like that.
.
Can be separated into section : 
- Color (Adjust place and render type)
- Font (Adjust place and size)
- Shadow

andy840119:
Got a idea
Maybe just throw playfield into here?
And should able to close playfield.

PR #225, opened 2020-10-24
Implement style editor right side

andy840119:
Implement part of #223 

image
This is how it looks like 👍


andy840119:
Also should think about make note style into separate editor or just in here?

andy840119:
Hmm...
Just make fast merge and think about note editor's part first.

Issue #215, opened 2020-10-17
Implement drag .lrc/.kar file to import.


andy840119:
Enable user to drag `.lrc`(same as `.kar`) file to import.

andy840119:
Test needed.
~~After test this issue can be closed.~~
This test cannot be tested because lazer client cannot support customize beatmap, sad.

andy840119:
Just close this issue now because it's implemented.

PR #230, opened 2020-10-26
Implement drag .lrc/.kar file to import. 

andy840119:
Implementation of #215.

todo : 
- [ ] Should show an alert to notify user this action will reset beatmap's all object(might create import nanager to deal with import logic)?
- [ ] Test in real client.


andy840119:
Not run that stupid test in appveyor.

andy840119:
This issue should be post-process until `ImportManager` has been implemented.

Issue #239, opened 2020-11-07
Enable to record which singer in lyric object.


andy840119:
For now we use `FontIndex` property to record Style index.
But eventually this property to record list of singer.
And might use this singer to find this lyric's skin.

List of Workable way : 
- Record list of singer index into this property.
- Record a number, and this number can convert to list of singer's id.

andy840119:
Needed in #214

andy840119:
2. is better, but need to think about the naming.

...
Maybe just named `ShiftingSingerIds`?

andy840119:
![image](https://user-images.githubusercontent.com/9100368/98456034-a83b6d80-21bb-11eb-9416-28f5d4fed853.png)

But `method1` is more reason and understandable 🤔 
And it's still can save list of singer ids into single number by using `JsonConversion` attribute.

andy840119:
Maybe just named `Singers` or `SingerIds` in `Lyric` class?

Note : 
Can just use `Singers` because `BeatmapInfo` just use `Bookmarks` to save list of bookmark ids in `osu.Game`

andy840119:
Implemented.

PR #240, opened 2020-11-08
Save list of singer index instead of font index in lyric.

andy840119:
Implementation of #239


andy840119:
image
Now got 12 testing error in this change.
Trying to implement `SingerUtils` and fix that 👍🏼

andy840119:
Seems can be merged now 🤔

PR #244, opened 2020-11-12
Record fill time tags(include duplicated).

andy840119:
It might use json as future beatmap's format, so write json convertor to deal with this.

Also, add `RowTimeTags` to record all `timetags` because it might have duplicated `timetag` in `japanese lyric`.


andy840119:
Also, not sure need to create `Timetag` class or just extension class, like `TimeTagUtils`
to deal with sorting and creating to dictionary.
.

Hmm...
I guess using utils is better.

andy840119:
Oops 
Also need to mark that time tags's value is nullable.

andy840119:
Wait.
Should it be changed into `IReadOnlyList`?

andy840119:
Looks ok.
About converting `RowTimeTags` to exist `Timetag` can leave to another pr.

PR #250, opened 2020-11-18
Base implementation of lyric import screen.

andy840119:
Because don't let user to have chance to mess-up.
So need to make those things in order:
  1. Create lyric text.
  2. Add ruby/romaji(need to check which language this lyric is).
  3. Add time tag(need to add ruby before).

And singer drag cannot handle such complex step, so this PR is create a screen with multi step to handle.
All import `.txt`, `.kar`, `.lrc` file will show this import screen and let use to run all the step.

There are total five step in this import screen:
- Drag
- Set language to each lyric (force)
- Run auto ruby generator and preview(if Japanese lyric contains)
- Run time tag generator and preview.
- Finish

Also, this screen stack design will reference how multi room design.
image


andy840119:
image
Now it's able to switch to different screen now.

andy840119:
It's able to give it a merge now.
then implement each step with different pr.

PR #251, opened 2020-11-20
Implement Japanese time tag generator

andy840119:
Implement Japanese time tag generator.

todo : 
- [x] generate base class and base parameter.
- [x] generate and implement Japanese lyric's time tag generator.
- [x] test case with test every config setting.


andy840119:
Ignore `CodeFactor` because rider says 👌🏼 .

PR #254, opened 2020-11-22
Implement time tag tooltip and it's test case.

andy840119:
This is how it looks like : 
image
image


andy840119:
Ingore `CodeFactor`'s stupid suggestion.

PR #257, opened 2020-11-22
Implement drag screen in lyric import.

andy840119:
Implement first page of #250 
What's add: 
- Enable to assign target file while opening lyric import screen.
- Separate import lyric manager from exist import manager.
- Move exist `ImportLyricDialog` into `ImportLyric >> DragFile` and rewrite logic.
- Give every sub screen an different icon.
- Ask before switching to different stage.
- OK dialog will have an call-back.


andy840119:
Fat code.

andy840119:
And seems it's ok enough, so give it a merge 🆗 .

Issue #263, opened 2020-11-28
Auto-detect lyric's language.


andy840119:
in #262 , it might be more user-friendly if can auto-detect which language lyric is.

andy840119:
Seems can use [Lucene.Net.Analysis.OpenNLP](https://www.nuget.org/packages/Lucene.Net.Analysis.OpenNLP/) on language detection.
See : 
https://www.itread01.com/content/1547731818.html

andy840119:
https://opennlp.apache.org/docs/1.9.1/manual/opennlp.html
Maybe need to read this article and understand how `LanguageDetector` works.

andy840119:
Have [LanguageGuessingTest](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Tests.Analysis.Phonetic/Language/Bm/LanguageGuessingTest.cs) sample in [Lucene.Net.Tests.Analysis.Phonetic](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Tests.Analysis.Phonetic/Language/Bm/LanguageGuessingTest.cs)

Guess it's implemented with [Beider-Morse Phonetic Matching](https://stevemorse.org/phonetics/bmpm.htm) and not accurate at all.
But it's a direction and maybe can add some language support like `Japanese` and change compare order?
Change order can be like : 
1. Japanese(can detect with is there any kanji and hiragana or katakana)
2. Chinses(with weird text)
3. English(a-z)
4. .....

And can let user set their common mapping language if language are in `other`

andy840119:
This issue can be closed due to implemented by another package.

PR #264, opened 2020-11-28
Implement auto-detect lyric's language.

andy840119:
Implementation of #263 .

The final way is use this package called [language-detection](https://github.com/pdonald/language-detection), it provide a simple way to detect language.
But now it does not fully support Japanese lyric, should fix this dictionary to add all Hiragana and Katakana.


andy840119:
Make a merge and fix that after.
Also maybe should make a way to import customize dictionary.

andy840119:
http://mail-archives.apache.org/mod_mbox/lucene-commits/201811.mbox/%3C8b9f623c2000452391fd671cb2fa7cb0@git.apache.org%3E
This is dictionary on define is Japanese text and has some error in here(e.g. : does not have `い`)
Need to find is there any newer version.

PR #270, opened 2020-11-29
proofreading to karaoke

zoogies:


andy840119:
nit : 
it good to check in line comment(`//`) or function comment(`///`)
- should be lowercase or uppercase
- should add `.` at the end of sentence?

It's better to define it at the fist time and trying to follow it 👍🏼

osu.Game.Rulesets.Karaoke.Tests/Beatmaps/Formats/NicoKaraDecoderTest.cs
```diff
@@ -19,7 +19,7 @@ public class NicoKaraDecoderTest
     {
         public NicoKaraDecoderTest()
         {
-            // It's a tricky to let osu! to read karaoke testing beatmap
+            // a trick to get osu! to register karaoke beatmaps
             NicoKaraDecoder.Register();
```
andy840119, line 23: maybe `a trick way`?

osu.Game.Rulesets.Karaoke/Edit/Generator/TimeTags/Ja/JaTimeTagGenerator.cs
```diff
@@ -33,7 +33,7 @@ protected override void TimeTagLogic(Lyric lyric, List<Tuple<TimeTagIndex, doubl
 
                 if (char.IsWhiteSpace(c) && Config.CheckWhiteSpace)
                 {
-                    // 空白文字の連続は無条件で無視
+                    // 空白文字の連続は無条件で無視 (Unconditionally ignore a series of whitespace characters)
```
andy840119: Can just remove Japanese comment.

andy840119 (commented):
Other LGMT.

zoogies:
> nit :
> it good to check in line comment(`//`) or function comment(`///`)
> 
> * should be lowercase or uppercase
> * should add `.` at the end of sentence?
> 
> It's better to define it at the fist time and trying to follow it 👍🏼

I think going based off that I see online for formatting it would be best to do all lowercase with no "." but its up to you how you want to format it

andy840119:
In lazer project, seems more of comment use Lowercase with "."
BTW, is there any article recommend using lowercase i can see?

zoogies:
Let's do lowercase with "."
I couldn't find an article so we should just do the same as lazer

andy840119:
I guess it's better to : 
```
// uppercase and lowercase is ok in single comment.
```
In single line case.

```
/// <summary>
/// Uppercase is required.
/// </summary>
```
In (i'm not sure it's called multi line comment) case.
All comment should with "."

BTW still need to do with this PR : 
- [ ] Check format the `exist file` you changed if you can(might be better even the comment you not modified before.)
- [ ] Remove that Japanese comment.
- [ ] Update or rebase to master to fix the conflict.

andy840119:
Other part LGMT

Susko3:
FYI the comments that begin with three `///` are called [XML documentation comments](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/), xmldoc for short.

For multi line comments that don't use the xmldoc format, you'd use `//`, or `/* ... */`.

Issue #275, opened 2020-12-01
English time-tag generator.


andy840119:
For now we have Japanese time tag in #251 
But it could be better to create some common language's time tag.
So user don't need to generate themself.

andy840119:
Also can create Chinese's one.

andy840119:
@Yoyolick You can try this issue if you want.
I will not implement it within a month.

zoogies:
I'll take a look 👍

andy840119:
Feel free to ask me if you are needed.(Even asking without watching the code is ok
Also recommend try to use some exist karaoke making software and make a karaoke format file to understand how it works 👍

andy840119:
https://stackoverflow.com/questions/10825685/chunking-english-words-into-graphemes-corresponding-to-distinct-sounds
Bookmark this urls for better searching.
.
keywords : `chunking  english words into graphemes C#`
Need to find is there any related package for that 🤔 ?

andy840119:
Make a memo : 
https://github.com/google/sentencepiece#end-to-end-example
It's a little bit like this project.
.
https://github.com/wolfgarbe/SymSpell
Not really sure can this project do the same things because it contains `word-segmentation` tag.

andy840119:
Not really sure but make a memo : 
https://www.nuget.org/packages/Stanford.NLP.Segmenter/
https://sergey-tihon.github.io/Stanford.NLP.NET//samples.html#Stanford-CoreNLP
Not really sure what this doing 🤔 .

PR #277, opened 2020-12-03
Create lyric creator for chinses lyric.

andy840119:
Implement part of #275 with chinses one.


andy840119:
Lazy to add other test case 😠

PR #278, opened 2020-12-03
update nuget package.

andy840119:
Update to latest version of `lazer`.


andy840119:
image
The only thing i can check is everything(?) works ok.

PR #284, opened 2020-12-05
Implement time tag display in lyric editor.

andy840119:
image
This is how it looks like 👍🏼


andy840119:
todo : 
- [x] need to delay time tag position calculation until `karaoke text` has been re-calculated.

andy840119:
It's able to give it a merge if CI says ok.

PR #287, opened 2020-12-06
Implement auto-generate in import screen.

andy840119:
What's add in this pr:
- Implement basic import feature(remove all hit objects and re-add)
- Rename from `LanguageManager` to `TranslateManager`.
- Implement ruby/ romaji manager and apply on import screen.
  - Will deal with create/delete ruby/romaji
  - Auto generate ruby/ romaji
- Implement time tag manager and apply on import screen.
  - Will deal with all time tag related.
- Implement lyric manager and apply on import screen.
  - Add auto-detect language support.
- Fix the case if time-tag to dictionary pass null


andy840119:
Should give it a pass 😐

PR #292, opened 2020-12-07
Implement bindable dictionary

andy840119:
Because `Lyric.Translate` need bindable dictionary, so implement is needed.
For now seems most test case run success, so might make a merge then use it.


andy840119:
\XXL/

Issue #283, opened 2020-12-05
Create language selection pop-up


andy840119:
Because has many language that is hard to handle it with dropdown.
So maybe create something like `LanguageSelectionPopUp` to handle that.

![image](https://user-images.githubusercontent.com/9100368/101235829-2fbec280-370f-11eb-8ede-3ef0f4de48e1.png)
Reference this.

![image](https://user-images.githubusercontent.com/9100368/101235836-3fd6a200-370f-11eb-955d-045e7206fe4d.png)
And this.

andy840119:
And should separate two parts:
- Common
  - In editor setting, can make some language as common selection.
  - All language use in this beatmap can place into here also.
- All
  - All available language.

andy840119:
And now seems only lyric maker will use this pop-up
Translate screen not sure.
Maybe can place into more common place.

andy840119:
Maybe has interface 
```
public virtual Bindable<CultureInfo> Current { get; set; }
```
Like other config control.

andy840119:
This issue can be closed due to implemented.

![image](https://user-images.githubusercontent.com/9100368/101661020-21fc9a80-3a8b-11eb-8b59-44769d6023f3.png)
And how it looks like :

PR #293, opened 2020-12-08
Use calture-Info instead of language id in translate.

andy840119:
Use calture-Info instead of language id in translate.

What's add in this pr : 
- Replace all `BeatmapSetOnlineLanguage` into `CultureInfo`
- Test case still pass, happy.

In fa9d93a3f484a3d1664d2819bf6f46009eb07cbb, need to fix:
- [x] Some language is commented out. (open issue in #295 )
- [x] Should move Language pop-up into another place, maybe manage list of layouts?
- [x] Should save calture id instead of string. (open issue in #294 )

And what's need to be added:
- Language selection dialog #283, will be used in default display language in setting pane.
- language preview pop-up can display language name now(if have.)


andy840119:
Maybe wait until #283 has been implemented?
It might less broken time in master.

andy840119:
Seems need to merge before #296

Issue #294, opened 2020-12-09
Recording default language as CultureInfo in config manager.


andy840119:
Because in #293 changes translate language recording object  from `Object`(And use it's language code as comparing prefer language) to `CultureInfo`.

So might need to deal the case : 
```Csharp
config.GetBindable<CultureInfo>(KaraokeRulesetSetting.PreferLanguage)
```

Need to think about how to deal with save/load from database.

andy840119:
This issue can wait until #283 because now `PreferLanguage` is not working.

andy840119:
This issue can be closed due to implemented.

PR #296, opened 2020-12-09
Implement translate selector.

andy840119:
Implementation of #283 .
Also deal with issue #294, now it's able to save/load language and can just use `config.GetBindable<CultureInfo>(KaraokeRulesetSetting.PreferLanguage)`.

image
This is how it looks like now

image
Also need to fix this in the future, maybe it's better to open language selector in the new screen.


andy840119:
Let's check is...
Oops CodeFactor is not happy.

andy840119:
Seem cannot be fixed today.

PR #297, opened 2020-12-10
Clean up code

andy840119:
Just clean-up code.


andy840119:
Cool
Seems it's ok to make a merge.

Issue #290, opened 2020-12-06
Implement lyric editor cursor.


andy840119:
Should be able to move cursor by clicking lyric text or time-tag.

andy840119:
Step : 
1. Might think about namespace adjusting.
2. Maybe has bindable property to record which time tag cursor in.
  - Mark a note that cursor only can in the place with time-tag.
3. Implement cursor 
  - Not really sure how should it like, rectangle or triangle is ok.
4. Deal with start and end case
5. Deal with the case that if cursor is in second duplicated position time tag.
6. Auto move to next cursor if call set time function().

andy840119:
This issue can be closed due to implemented.

PR #299, opened 2020-12-10
Implement cursor in lyric editor.

andy840119:
Implementation of #290.

What's add in this pr : 
- [x] Have cursor object
- [x] Display at the screen.
- [x] Can move left an right.
- [x] Can move up and down.
- [x] Appear at the first index.


andy840119:
Found the bug that use `Tuple` will cause cursor position compare error.
Should replace from `Tuple<TimeTagIndex, double?>` to struct `TimeTag`

andy840119:
Cursor moving issue has been fixed.

andy840119:
The remain things is define moving up and down's position.
Also should have feature to click the time-tag to move.

andy840119:
Still cannot be triggered.
Seems like it's caused by placing `DrawableTimeTag` into internal.

andy840119:
Although time-tag cannot be click now, but it's ok to give it a merge now.

PR #304, opened 2020-12-12
Press space to set current time in time-tag

andy840119:
What's add in this pr:
- Set time-tag's time.
- Clear time-tag's time.
- Add time-tag.
- Remove time-tag.


andy840119:
Just ignore large functions issue.

andy840119:
Might have little bug but can fix after.

Issue #259, opened 2020-11-24
Lyric editor's mode.


andy840119:
Due to most of time user will manage their lyric related things in this page, so need to define list of mode here.

Lyric edit mode : 
- View mode
  - User cannot edit anything except each lyric's left-side part.
- Edit mode
  - Can create/delete lyric
  - Can move lyric's position
  - Can split/combine lyric
- Record mode
  - Click white-space to set current time into time-tag.
- Time tag edit mode
  - Enable to create/delete time tag.
  - Reset single time tag's time.
  - (optional)Hover to see time tag's time.
  - Show hidden time-edit panel while click the lyric.

Left side edit mode : 
- Time and layout.
- Time and language.
- Time and singer.
- Time only.

andy840119:
This issue will un-pin until switch mode has been implemented.

andy840119:
This issue can be unpin now.

andy840119:
This issue can be closed because main part has been implemented.

PR #308, opened 2020-12-13
Implement lyric utility to handle split and combine lyrics.

andy840119:
In order to implement edit mode in lyric editor (#259)
Need to implement utility to handle split and combine lyrics.


andy840119:
Ignore large function and `()`

andy840119:
Seems it's ok to give it a merge.

PR #313, opened 2020-12-13
Implement note utility to handle separate and combine notes.

andy840119:
https://github.com/karaoke-dev/karaoke/blob/edbb0de1465625fa1f030f0370c6098baee4db8e/osu.Game.Rulesets.Karaoke/Objects/Note.cs#L102

Seems it's not good to writing separate note logic into here, weird and hard for testing.
So moved it into `Utils` namespace and writing like other utilities.


andy840119:
Got no complain in code-factor, good.

andy840119:
Other part lazy to write tests now 😮

Issue #315, opened 2020-12-15
Create state manager into lyric editor


andy840119:
Duty of this manager:
- mouse position
- edit mode
- left-side mode

andy840119:
Final name is called `LyricEditorStateManager`.
Named `XXXManager` is ok, but not inherit `Component` because this manager should only record state. should not access drawable part.

andy840119:
This issue can be closed due to almost implemented.

PR #316, opened 2020-12-15
Create state manager into lyric editor

andy840119:
Implementation of #315.

What's should be added in this PR : 
- [x] Add manager
- [x] Move cursor position state into manager.
- [x] Move mode state into manager.
- [x] Implement moving tag cursor in edit tag mode (`EditMode`).
- [x] Implement moving cut cursor in lyric edit mode (`TimeTagEditMode`).
- [x] Able to switch mode in test case.


andy840119:
Notice that it might need to handle select range of text in new `edit lyric text` mode in #321 
So it might make a merge now and throw remain into another pr?

Issue #320, opened 2020-12-17
Enable to add or delete text in lyric utils


andy840119:
It should be able to:
- add text at the target position.
- delete range or text on target position.

andy840119:
Can be closed due to implemented.

PR #322, opened 2020-12-18
Enable to add or delete text in lyric utility

andy840119:
Implementation of #320


andy840119:
I'm not sure but is appveyor run faster recently 🤔 ?

PR #330, opened 2020-12-20
Add shifting in utils.

andy840119:
What's add in this PR : 
- Add `TimeTagIndex` shifting in utility.
- Add `TimeTag` shifting in utility.
- Add `TextTag`(ruby and romaji tag) shifting in utility.
- Add missing time-tag process while add or delete text in lyric utility.

Also has test case in those utils.


andy840119:
Seems CodeFactor sick today.

Issue #333, opened 2020-12-20
Implement romaji auto generator


andy840119:
Step about how to implement : 
 Use [HiraKana](https://github.com/Xifax/HiraKana) package to convert `Hiragana` to romaji.
  - Add `ToRomaji` and `RomajiToHiragana` in `JpStringUtils`.
  - Add Unit test about it.
  - Create `JaTimeTagGenerator` in `Generator.RomajiTags.Ja`.
  - Add unit test about it.

andy840119:
Or using https://github.com/caguiclajmg/WanaKanaSharp because it's on nuget?
But it does not support convert romaji to Hiragana

andy840119:
Or test this project:
https://github.com/Cutano/Kawazu

Seems it also able to convert kanji to Hiragana?
Oops it's too fat.

andy840119:
Also should throw exception in helper if tag format is not valid

andy840119:
Cal be closed due to implemented.

PR #338, opened 2020-12-21
Implement Japanese lyric romaji generator.

andy840119:
Implementation of #333


andy840119:
Guess make a merge now 🤔 
Also might have an option to auto-generate romaji if not have in beatmap.

andy840119:
Oops

PR #345, opened 2020-12-26
Separate auto-generator selector.

andy840119:
What's add in this pr:
- Ruby auto generator selector
- Romaji auto generator selector
- Time tag auto generator selector
- Skip ruby/romaji step if not generatable.
- Enable to generate romaji in ruby/romaji generator step.


andy840119:
Ignore what appveyor says.

PR #348, opened 2020-12-27
Fix lyric not display in singer editor.

andy840119:
This is the demo result on why lyric not display in timeline.
Need to have a better coding style.
.
And this PR will be process until it's able to switch page in editor.


andy840119:
Make a note: 
might adjust pooling size to let more lyric blueprint into preview area.

andy840119:
Fix the conflict.
Note that should make a rebase until merge.

Issue #360, opened 2021-01-01
Auto layout generator


andy840119:
Maybe it's necessary to design layout pattern to let user able to auto-generate lyric layout position.

andy840119:
Maybe add `Group` and `Order` property in `KaraokeLayout`?

andy840119:
And it's able to store some default layout in ruleset, so lyric arrangement is workable even it has no layout store in the beatmap.

andy840119:
And the implement step:
- Rename component from `Skinning/Components` into `Skinning/Metadatas`, also separate `Layout` and `Style`.
- Maybe separate `.skin` file into `.layout` and `.fonts`?
- Create some `.layout` files to store list of groups and default layout?
- Add `Group` property to store group name.
- Add `GroupGenerator` it's able to generate list of layout by selecting list of lyric.

andy840119:
Also need to think about mapping lyric layout to local or skin layout.
.
Maybe it's able not to think about it because it doesn't have any karaoke skin resource in beatmap now.
And means step 2 and 3 can be ignored, just need to add extra property in skin.json now.

andy840119:
Also might think about how to get skin resource in generator.

andy840119:
And how to interact with generator in lyric editor?

andy840119:
Basic lyric generator has been implemented.
So might ok to close this issue now.

PR #363, opened 2021-01-02
Auto layout generator

andy840119:
Base implementation of #360 
.
What's add in this pr:
- Implement base generator (add test case also).
- Add sample group in resource.
- Able to convert string to lyric in test case helper.


andy840119:
The full-implementation will be wait until this generator has fully in production.

PR #364, opened 2021-01-02
Switch lyric edtir in main editor

andy840119:
For now finally it's able to switch `LyricEditor` or `EditPlayfield` in karaoke editor.

And what's change in pull request;
- [x] should not display setting button(overlay) in editor page.
- [x] has mode enum.
- [x] has bindable to record mode(place in composer)
- [x] apply lyric editor.
- [x] able to switch mode(include UI and input part, and should add button to handle it now).
- [x] should have its own clock in lyric preview (not really sure just throwing `GameplayClock` or make lyric editor's own clock).
- [x] check is other test case broken(Import and lyric editor screen page, and should remove duplicated time injection).
- [x] Move lyric editor screen into test case or just remove because it will not become an individual screen page.


andy840119:
Thanks seems has ok place to add lyric editor as main editor.

Issue #365, opened 2021-01-03
Design editor menu bar


andy840119:
Although it not able to let customize ruleset to implement it's own editor menu bar, but it's able to planning it in test case.

Might have menu like : 
- File
  - Import from text.
  - Import from .lrc
  - Export to .lrc
- Edit(Lyric editor)
- Edit(Lyric)
- Edit(Note)
- Timing
- View
  - Edit mode
    - Lyric editor
    - Playfield
- View(Lyric editor)
  - font size
  - edit mode
  - left side mode.
- View(Playfield)
  - ??
- Tools
  - Singer manager
  - Translate manager
  - Layout manager
  - Style manager
- Options
  - Lyric editor
  - Generator
    - Ruby
    - Romaji
    - Time-tag
    - Layout
    - Auto-detect language

andy840119:
Also might make something like `KaraokeEditRulesetConfigManager`
It's ok to create and just use in this test case only.

andy840119:
Guess this issue can be closed now.

PR #367, opened 2021-01-04
Implement editor menu bar

andy840119:
Before waiting for lazer to enable customize ruleset to access menu bar, it's a choice to implement some basic feature.
And this is the implementation of #365 
.
What's add in this PR:
- [x] Add all selections listed in #365 
- [x] Enable to read/write settings from `KaraokeRulesetEditConfigManager`.


andy840119:
Also should not assign name inside class ctor.

Not like : 
```
public WaveformOpacityMenu(OsuConfigManager config)
            : base("Waveform opacity")
```

should like : 
```
new WaveformOpacityMenu("Waveform opacity")
```

Because of style issue and not really sure what will multi-language will works like(it's hard to switch language in ctor).

andy840119:
Guess it should pass test case 🤔

PR #374, opened 2021-01-07
Switch tooltip in edit lyric step in lyric importer.

andy840119:
todo : need to fix `ScheduleAfterChildren` in `LyricEditor` will cause get property delay.


andy840119:
Cool.

Issue #377, opened 2021-01-07
Editor config visualization.


andy840119:
It's time to let user able to make a config.
And because it might have lots of space only have setting(and most setting are bool type).
.
So it might better to add preview in config also.
.
![image](https://user-images.githubusercontent.com/9100368/103904955-53778d00-5141-11eb-8891-3240715c7eef.png)
Left side is config, and right side is preview area.
And it's able to let user know what actually change if user change the selection.

andy840119:
Separate two part in right side.
Upper part is the generate result by default, and botton is result generate by change value.
Also preview only works if user hover to the selection.
and preview will be changed if user trigger the change if config is not bool type.

andy840119:
And think about how to deal with those generate config visualization : 
- LanguageDetector(can skip in this pr)
- LayoutGenerator(can skip in this pr)
- RomajiTagGenerator(needed)
- RubyTagGenerator(needed)
- TimeTagGenerator(needed)

andy840119:
Or can just create config without preview feature (but this feature will be opened until preview has been implemented).
And make `TestSceneGeneratorConfigDialog` test case for opening all the dialog.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/103906717-a8b49e00-5143-11eb-91a7-098437825708.png)
And make a note that it's setting page from lyric maker.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/103906907-e285a480-5143-11eb-8f3b-56f2be1b3d63.png)
And this is setting from lyric editor.
Make a mote also.

andy840119:
And in this config it need to have : 
- Change color in different category of config dialog
- Change title and description.
- Add ok, cancel, reset button.
- Implement other category(ruby and romaji tag).

andy840119:
It will spend too much time on the visualization part, and config name and config UI element (e.g. slider, dropdown) should be enough.

PR #378, opened 2021-01-09
Config visualization.

andy840119:
This is the fist step of #377.
Only have config, preview part haven't implemented.
.
Also has rule that centralized save, load, default value and disable logic.


andy840119:
Ignore what appveyor says.

Issue #401, opened 2021-01-13
Enable to import lyric in editor screen.


andy840119:
Enable to import lyric in editor screen.

Where to import:
- Import from menu bar.
- Import from drag file to screen(Cannot tested it now so ignore implementation.)

Also make a note: 
- Should remove drag file from lyric editor because it might cause import recursive.

andy840119:
Cannot implement now because it require switch to new screen.

andy840119:
Also should have pop-up to select file first.

andy840119:
Make a note : 

Can use `RegisterImportHandler` in `osuGame` to import some dragging file can be imported,
For example : 
- LrcImportManager

andy840119:
Also need to check why `ImportLyricManager` is in here and where is fit to place it.
It looks weird now.

andy840119:
need to think about it and re-make a step design about it.

![image](https://user-images.githubusercontent.com/9100368/104467448-acd83400-55f9-11eb-9226-58e43b1838ed.png)
And should think about using `FileImportScreen` or not?
(seems it's not able to assgn target format of file, so might need to create a new one 🤔

andy840119:
Just able to drag `.lrc` or `.txt` file only in first import screen.
Over.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/104682011-0db95680-5737-11eb-953e-39477e63c559.png)
Make a screenshot
This is what before looks like

andy840119:
This issue can be closed due to implemented.

PR #402, opened 2021-01-15
Implement import step's selecting file.

andy840119:
Implementation of #401 

image
Before

image
After

What's add in this pr: 
- Have file selection in this page.
- Define that only this page can process dragging to import event.
- Do no initialize time-tag if already contains.


andy840119:
Just ignore this shit suggestion.

PR #408, opened 2021-01-16
Refactor input manager.

andy840119:
What's change in this PR:
- Make saiten input manager and contorl input manager into individual file.
- Make saiten input's order to first (for better understanding)


andy840119:
Make a merge and add input manager for editor one 🚀

PR #415, opened 2021-01-17
Add area to trigger add singer

andy840119:
image
Add a stupid button below to singer list.


andy840119:
After this pr, need to think about the view to fill new singer.

andy840119:
Link issue #416 for better tracking.

Issue #416, opened 2021-01-17
Tracking board to implement singer editor.


andy840119:
What's still need to be implemented in singer editor:
- Create singer (in #416).
- Delete singer (in #416).
- Right-click to edit/delete singer (in #416).
- Fix lyric disappear issue (in #348  )
- Able to edit singer info (in #427).
- Able to update singer id if singer change the position. (in #418)
- Create logic in manager. (implemented)
- Fix lyric not appear issue. (will implemented in #348 )

andy840119:
Also notice that might able to add or remove singer index in utils (lyric utils).

andy840119:
And not really sure should move `assign singer index` into `lyricManager`.
If that so, then might need to make a refactor.
.
Update : got no  idea what i said before but seems this comment has been implemented.

andy840119:
Guess most of parts has been implemented.
This issue seems can be closed.

PR #418, opened 2021-01-17
Add add/remove singer utils.

andy840119:
What's add in `LyricUtils`:
- Add singer to lyric.
- Remove singer from lyric.
- Check is singer in lyric.
- Check lyric's singer singer list is in target singer list.


andy840119:
Linked to #416

Issue #413, opened 2021-01-17
Update to net5


andy840119:
https://github.com/ppy/osu/pull/10926
See this issue and make a copy-paste

andy840119:
Update works.
Need to test is github action also works 🤔 ?

andy840119:
https://github.com/karaoke-dev/karaoke/runs/1717134183
Ass it's not working.
Test it in release build on local then try again.

andy840119:
Oops runs well on my laptop.
Guess should test it on my stupid mac.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/104922128-9fe28880-59dd-11eb-9f14-90e237ffff79.png)
Seems it's caused by `ILRepack` in linux.

andy840119:
Seems it's caused by `ILRepack` in linux.
https://github.com/gluck/il-repack/issues/273

andy840119:
https://github.com/Game4all/gamebosu/commit/d17412f395395ec73c0235179267737b0bd3019c
Or just use fody instead?

andy840119:
Still working

here's the guide of `ILMerge.Fody`
https://github.com/tom-englert/ILMerge.Fody 

and here's the packing list 
https://github.com/karaoke-dev/karaoke/blob/master/osu.Game.Rulesets.Karaoke/osu.Game.Rulesets.Karaoke.csproj

andy840119:
use fody seem will cause `The type 'XXX' exists in both ...` error
Maybe try to use `ILMerge` instead 🤔 
https://blog.walterlv.com/post/merge-assemblies-using-ilmerge.html
https://github.com/dotnet/ILMerge

And it works more like `ILRepack`
here's the document:
https://github.com/dotnet/ILMerge/blob/master/ilmerge-manual.md

andy840119:
Summarize the result:
- exist ILRepack: not working well net5
- fody ILRepack: will override `osu.game.ruleset.karaoke.dll`, which will cause ` The type XXX exists in both ...` error.
- ILMerge: cause error if pack `osu.game.ruleset.karaoke.dll`.

andy840119:
The laziest way is just roll-back to .netcore3.1, but need to re-add ` global.json`
Or release `.dll` by hand recently.

andy840119:
or just let `ILMerge` pass because not make release build in project 🤔

andy840119:
[...]

PR #419, opened 2021-01-17
Update nuget package.

andy840119:
Implementation of #413
Update nuget package.
Also upgrade target to .net5.0


andy840119:
https://github.com/ppy/osu/commit/7c612ec5561225b2a618bf11915aa693e20e7cf6
not really sure why remove `global.json` 🤔

andy840119:
anyway it works 🎉

PR #425, opened 2021-01-23
Display singer avatar in lyric editor.

andy840119:
Display singer avatar in lyric editor.
This is initialize implement and might need UI improvement.


andy840119:
Can make a merge now.
Might need to add right-click selection for it.
But need to move tooltip generator from `SngerManager` into `LyricManager`
~~Or just create tooltip for that?~~

PR #427, opened 2021-01-24
Implement edit singer info

andy840119:
Implementation part of 416

What should be done in this pr:
- [x] pop-up singer info.
- [x] enable to type singer info.
- [x] enable to write-back singer info.
- [x] update text after edit (or close) singer edit pop-up.


andy840119:
Got a big problem:
Seems `TriggerChange()` only trigger in itself and setvalue not wotking.

andy840119:
Use tricky way to deal with that shit

andy840119:
Anyway it works.
Make a merge and watch animate.

Issue #430, opened 2021-01-26
Cursor skip mode in recording time-tag


andy840119:
Should have those options:
- None
- SkipStartTag
- SkipEndTag

In RecordingMovingCursorMode

andy840119:
And should be visible which is movingable in lyric editor

andy840119:
Notice that need to place to config in editor also.

andy840119:
This issue can be closed due to implemented.

PR #432, opened 2021-01-26
Cursor skip mode in recording time-tag

andy840119:
Implementation of #430 .
What's should be added in this pull-request:
- [x] Add enum.
- [x] Enable to setting in config and menu.
- [x] Implement moving logic.
- [x] Should be visible which is movingable in lyric editor (might only need to change in recording mode).


andy840119:
image

\Almost there/

Issue #433, opened 2021-01-28
Implement create/delete/combine/separate lyric.


andy840119:
In [LyricManager](https://github.com/karaoke-dev/karaoke/blob/master/osu.Game.Rulesets.Karaoke/Edit/Lyrics/LyricManager.cs#L20) and [TimeManager](https://github.com/karaoke-dev/karaoke/blob/master/osu.Game.Rulesets.Karaoke/Edit/Lyrics/TimeTagManager.cs#L22), we use `EditorBeatmap` to get `list of lyric` directly.
But it might be hard to deal with the case with `create`/`delete` lyric (especially in `LyricManager`)

One of the way to deal with this issue is use `BindableList<Lyric>` in `LyricManager` and `EditorBeatmap`, then bind with in `LyricEditorStateManager`.

andy840119:
And should initialize `BindableList<Lyric>` in `LyricEditorStateManager`
Deal with create/delete saving event in `LyricManager`(or LyricEditorStateManager)
Not really sure about that 🤔 
.
[Edit]
Also need to think about where to deal with cut or combine lyric 🤔

andy840119:
Also need to able to right-click to remove lyric.
And have button to click to add lyric below.

andy840119:
This issue will be implemented after #434

andy840119:
Right-click should have:
- Delete lyric
- Combine with previous.

And right-click only works in edit lyric mode.

andy840119:
And might fix moving cursor with order with same PR.

andy840119:
This issue can be closed now due to feature has been implemented.

PR #436, opened 2021-01-28
Use bindable to get list of lyric

andy840119:
Implement issue #433 

What's should be added:
- [x] use bindable list.
- [x] moving lyric cursor should be ok after change lyric order.
- [x] should sync display way with singer editor.
- [ ] able to create/delete lyric in editor.
- [ ] able to split/combine lyric in editor.
- [ ] right-click to combine/delete lyric (only in lyric edit mode).


andy840119:
Most feature has been implemented.

What's remain : 
- Combine lyric.
- Test create/delete/split lyric because it'll cause crash issue created in drawable lyric text.

andy840119:
And ignore what code factor says,
Just let it pass.

Issue #389, opened 2021-01-09
Able to lock move in lyric editor.


andy840119:
Able to lock move in lyric editor.

andy840119:
Changing lyric editor order only works in `EditMode` and `TimeTagEditMode`

andy840119:
Can be closed due to implemented.

PR #442, opened 2021-01-30
Able to lock change order in lyric editor.

andy840119:
Implementation of #389 
Now it's able to disable dragging if not in edit mode in lyric editor.


andy840119:
Seems can merge this cool stuff 🥳

Issue #459, opened 2021-02-10
Adjust lyric editor color.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/107494870-46c7d800-6bca-11eb-8ea5-7c0aebf69445.png)
Here's the expect change.
It should be better to change color follow what editor have.
Mark the hex code or name in osuColour if have enough.

andy840119:
editor color : 
- Left side time area : Color4Extensions.FromHex("333") in `TimelineArea.cs`
- Left side content area: Color4Extensions.FromHex("222") In `TimelineArea.cs`
- Left side static area : colours.Gray4 in `BeatDivisorControl.cs`
- Right side lyric area: Color4.Black with alpha = 0.5f in `PlayerSettingsGroup.cs`

andy840119:
This issue can be closed due to implemented.
Right-side static area can be ignore now because not sure will apply or not.

PR #462, opened 2021-02-11
Apply color change.

andy840119:
Implementation of #459 

image
Before

image
After


andy840119:
Guess it's ok to make a merge.

Issue #455, opened 2021-02-08
Able to right-click to create new in edit lyric mode.


andy840119:
Should have right-click to create new in edit lyric mode.
For insert new lyric below target lyric.

andy840119:
This issue can be closed due to implementing.

PR #463, opened 2021-02-11
Able to right-click to create new in edit lyric mode.

andy840119:
Implementation of #455 
.
TODO : 
need to test it on my PC.
And  because it will cause crash issue in `osu.Game.Rulesets.Karaoke.UI.Components.SaitenVisualization.Add`
So not able to test now 😢


andy840119:
Also should add option to right-click to combine previous lyric(first lyric should not shows this option)

andy840119:
Ignore test and let this shit pass.

Issue #412, opened 2021-01-16
Show alert checker in lyric editor.


andy840119:
Might have a place to should all exist error in each lyric line.

All error might have:
- Invalid lyric time : update_time
- Invalid lyric time compare with other : update_time_tag, update_time, other_time_event,
- Invalid time-tag time : update_time_tag
- Ruby out of range : update_ruby
- Ruby duplicated position : update_ruby
- Romaji out or range : update_romaji
- Romaji duplicated position : update_romaji
- Missing translate : update_translate

andy840119:
Might update error list if figure out.

andy840119:
And might use flag(enum) to organize the error?

andy840119:
Also it will be better if can shows the position.
If that so, should deal with utils layer.

andy840119:
BTW, it might find invalid by utils, but where to handle?
Or should make something like `InvalidCheckerManager` ?

andy840119:
Also should record the result into some place 🤔 ?

andy840119:
Also is there any way to cache the result and can be injectable 🤔 ?
Like it will be calculated  and store in `LyricEditorStateManager`.
But it might be a pain to update view .
.
Need to think about where to update and where to send notification.

note : 
might able to use `beatmap.HitObjectUpdated` but should call `Update` while object updated.

andy840119:
And if wants to update invalid time-tag visualization, where to calculate that shit 🤔 ?
Notice that calculate algorithm might need editor config.

andy840119:
Maybe just need to calculate `invalid` lyric in LyricEditorStateManager, invalid from lyric itself should update from component itself?

andy840119:
Remain things in this issue:
- [x] Tooltip should be able to target position.
- [x] Tooltip should be more colorful(might create `InvalidMessageTextFlowContainer` with method to handle that with `OK`, `Warning`, and `Alert` state)
- [x] Adjust check icon position.
- [ ] apply auto-fix by right-click (notice that tooltip and right-click context menu cannot appear at the same time).
- [ ] Update invalid style in lyric.
- [ ] Should trigger update check after lyric property changed(need to think about where to call).

andy840119:
This issue is too old and now lyric editor can:
1. Show the errors by lyric(e.g. hover to alert icon at the left side of the lyric and it can shows the error with this lyric).
2. Show the error by edit mode(e.g. all error that caused/can be fixed in the edit time-tag mode.)

PR #469, opened 2021-02-15
Show alert checker in lyric editor.

andy840119:
This is the first implementation of #412 

What's been done in this PR:
- [x] Check listed invalid property in #412
- [x] Enable to find invalid reason by flag.
- [x] Has report to record checking result.
- [x] Enable to call for update checking(and support partial update).
- [x] Test case for checking class.
- [x] Have tooltip(alert message in this PR).
- [x] Place this icon into lyric editor. 

What's should be done in next PR:
- Update invalid style in lyric.
- Should trigger update check after lyric property changed(need to think about where to call).
- apply auto-fix by right-click.
- Tooltip should be able to target index.
- Tooltip should be more colorful(might create `InvalidMessageTextFlowContainer` with method to handle that with `OK`, `Warning`, and `Alert` state)
- Adjust check icon position.


andy840119:
This issue can be merged after CI because it's toooo fat.

Issue #457, opened 2021-02-08
Enable to auto move scrolling position while change selected time-tag position.


andy840119:
The main goal wants to change the  scrolling position is: 
- Change time-tag cursor by keyboard.
- Recording mode change time-tag.
.
Need to check is it be ok if use click the time-tag and position has been changed also?
And should have config that moving should be in which line?
And focus target should be `recycling container item`, not time-tag itself.

andy840119:
This logic should works if not in view mode.
Current edit lyric style is handle in #456

andy840119:
And mouse change time-tag position or cursor position is also works.

andy840119:
This issue can be closed due to implemented.

PR #474, opened 2021-02-16
Auto move scrolling position while change selected lyric.

andy840119:
Implementation of #457


andy840119:
Note : 
```
osu.Framework.Allocation.DependencyNotRegisteredException: 'The type LockInfo has a dependency on KaraokeRulesetEditConfigManager, but the dependency is not registered.'
```
It might cause dependency is not registered error.
Not cause in this PR but night need to fix after.

Issue #479, opened 2021-02-16
create InvalidMessageTextFlowContainer to show alert message.


andy840119:
Part of issue #412 
Need to display those type of message:
- OK
- Warning
- Alert

All message should display with prefix icon and target color.
And icon / text color might different.

andy840119:
Also should add test case.

andy840119:
Or called `WarningTextFlowContainer`?
Need to think better name.

andy840119:
This issue can be closed due to implemented.

Issue #475, opened 2021-02-16
Fix some dependency not inject error in import lyric page.


andy840119:
```
osu.Framework.Allocation.DependencyNotRegisteredException: 'The type LockInfo has a dependency on KaraokeRulesetEditConfigManager, but the dependency is not registered.'
```
It might cause dependency is not registered error.
Not cause in this PR but night need to fix after.
.
Also need to think about how to inject checker in lyric editor?

andy840119:
Fixed.

PR #481, opened 2021-02-17
create message container to show alert message.

andy840119:
Implementation of #479 

image
This is how it looks like.

image
And in tooltip.


andy840119:
todo : has beautiful text in paragraph.
need to check how it works in framework.
Also need to think about should make function overridable(to let user easy to customize their text color.)

[Update] : 
able to add text after `AddSuccessParagraph`.

andy840119:
Also check able to adjust text spacing?

[Update]:
Has spacing property.

andy840119:
This issue can be merged after CI give it a pass.
Because this issue is focus on display message container, not message in invalid lyric.

andy840119:
Ignore this test case and fix it in #475 .

Issue #490, opened 2021-02-20
Make cursor calculation more abstract


andy840119:
Inspired from #488 
For now there's two algorithm for calculation cursor position in `LyricEditor`
But they might become more abstract, like how `IScrollAlgorithm` do.

Step : 
- make `GenericCursorPositionAlgorithm` and `RecordingCursorPositionAlgorithm`, and both of them has same interface(`ICursorPositionAlgorithm`).
- Create`GenericCursorPosition` and `RecordingCursorPosition`, and both of them has same interface(`CursorPosition`).
- Some common interface in `ICursorPositionAlgorithm`
  - bool ToTargetPosition(CursorPosition position);
  - bool PositionMovable(CursorPosition position);
  - CursorPosition GetPosition(MovingCursorAction action, CursorPosition  current)

andy840119:
Also might make a test for that.

andy840119:
This issue can be closed due to implemented

PR #491, opened 2021-02-20
Refactor cursor position algorithm

andy840119:
Implementation of #490 
What's add in this pr:
- [x] Some utils to check is time-tag or text tag out of range.
- [x] Make cursor position calculation algorithm more abstract.
- [x] Remove old logic.
- [x] Clean-up code in lyric editor(include better naming).
- [x] Should re-generate algorithm set if change some setting or create/delete lyric.
- [ ] Some test case to test algorithm.
- [x] Algorithm with edit(cannot hover to first or last index of lyric) and edit mode.


andy840119:
Lazy to write test case.

Issue #495, opened 2021-02-21
Make CaretPosition abstract


andy840119:
Idea is from #494 
Because there are three kinds of `CaretPosition` use the same class:
- TextCaretPosition
- TimeTagIndexCaretPosition
- TimeTagCaretPosition

Maybe should make an interface called `ICaretPosition`
And let each algorithm has it's own class.

andy840119:
Need to think about : 
- how to deal with `caretMovingAlgorithmSet` because it's hard to define common interface with template.
- need to change bindable value to different class if switch the mode.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/108623339-dd757e00-7481-11eb-9de2-55eb70d5576b.png)
Current issue: 
Got no idea how to pass type to template

andy840119:
This issue can be closed due to implemented.

PR #496, opened 2021-02-21
Make caret position abstract

andy840119:
Implementation of #495 

What's implemented in this PR:
- [x] Add `ObjectExtension`, which can call method by name.
- [x] Create `TextCaretPosition`, `TimeTagIndexCaretPosition` and `TimeTagCaretPosition`
- [x] Make each algorithm can accept it's own class.
- [x] Fix compile error.
- [x] Make code works. 
- [x] Maybe should give those class or algorithm better name.
- [x] Refactor caret.
- [ ] Might make caret position calculation more abstract.
- [x] Fix some feature might not working if object is `TextCaretPosition`.
- [x] Check is typing and edit algorithm works well.
- [ ] Writing tests in this pr?


andy840119:
Also might `move` those method from `interface` to `base class` because those method is called by name.
And inherit the interface from `base class`

andy840119:
Remain todo might not need to implement now.

Issue #355, opened 2020-12-29
Replace lyric editor with ruby/romaji editor.


andy840119:
I think the final goal is using lyric editor to edit ruby and romaji.
Need to think about how UI looks like and how to implement.
.
Will have list of todos below.
.
And what's need to be concerned about:
- how to insert ruby or romaji.
- how to move the range.
- how to edit text.
- how to remove text.

andy840119:
Maybe should add now mode called `ruby/romaji edit mode`?

andy840119:
And some extra need to be concerned:
- Should something exist blueprint like `OverlaySelectionBlueprint`?
- Select multi ruby/romaji, and how to define `select`?
- How to combine/separate ruby/romaji.

andy840119:
Old ruby/romaji editor removed in #633

Issue #503, opened 2021-03-01
Implement ruby/romaji caret position calculation algorithm.


andy840119:
It's the first step of #355 
Might have two kind of algorithm?
- Hovering/select exist Ruby/Romaji text tag.
- Typing caret for ruby/romaji text tag.

andy840119:
Need to think about the UX part, or the step on how to edit exist or create new ruby/romaji

andy840119:
And need to think how to create new ruby/romaji on lyric editor.

andy840119:
Also need to switch modes?
Like `edit ruby/romaji mode`?
And double click to edit ruby/romaji text?

andy840119:
Implemented.

PR #506, opened 2021-03-06
Implement ruby/romaji caret position calculation algorithm.

andy840119:
It's the first step of #355 
Implement moving cursor algorithm for that.
Test case are included.


andy840119:
Also is implementation of #503

Issue #157, opened 2020-08-02
Translate editor improvement.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/89119701-64193c80-d4eb-11ea-8acf-232eb66e7c32.png)
1. Add an option to close time.
2. [Optional] Enable to adjust column width.
3. [Optional] Pop-up an dialog to show save alert before switching to another language
4. [Optional] Add new... selection in language drop-down.
5. Add tooltip in Lyric to show fill lyric line.

andy840119:
This issue need editor re-write.

andy840119:
This editor can be implemented now.

TODO : 
Need to think about where to place?
Also think about where to place other editor like singer editor, ruby/romaji editor...

andy840119:
![image](https://user-images.githubusercontent.com/9100368/96201926-d224ba00-0f98-11eb-9af7-3be383fb4039.png)
Or just follow the official's one?
in `Edit/Translate`

andy840119:
Also think about how to make navigation?

andy840119:
Translate editor has been implemented.

Remain part : 
- [ ] Manage language(add, rename, resort and delete language).
- [x] Implement test for `LanguageManager`.

andy840119:
This issue should (or called `can`) be closed in this year.
No need to wait for editor.

andy840119:
Seems implemented in #350 
Translate editor can be seems as fully implemented.

PR #516, opened 2021-03-21
Adjust translate editor.

andy840119:
Should not able to type in textbox if contains no language.


andy840119:
Also ask only once if contains no language after switch to translate editor.

andy840119:
Mark that this is the improvement of #157

PR #532, opened 2021-03-31
Update nuget package.

andy840119:
Update nuget package.


andy840119:
Got no idea why CI not working.
Sad

Issue #542, opened 2021-04-04
Implement right-side preview component in karaoke setting subscription.


andy840119:
In each `SettingsSubsection`, it should have preview component can let user to see the instruction or preview the change.
And here's the list of component should be created eventually:
- Config
  - [ ] Common -> ???
  - [ ] Note -> show the note playfield with notes appear, maybe with one second one note. with random place?
  - [ ] Lyric -> show drawable lyric text.
- Style
  - [ ] Lyric -> with preview lyric, time is freeze, or time is adjustable?
  - [ ] Note -> with preview note playfield, time freeze.
  - [ ] Manage font -> might have instruction to show how to import font in right side, not sure should show all available font.
- Scoring
  - [ ] Device -> might display tone and microphone volumn in right side.
  - [ ] Pitch -> ???

PR #543, opened 2021-04-04
Implement base preview component.

andy840119:
Implementation of #542 
What's new in this PR:
- Implement base API for letting setting subscription can has its own component.
- Implement hover background effect for letting user know what is the preview area.
- Implement not implement component hint for default preview component.
- Adjust style.
- Prevent scroll event cause volume control appear.


andy840119:
Let this shit merge.

PR #550, opened 2021-04-08
Update nuget package.

andy840119:
update nuget package and fix some compile error.


andy840119:
Note : because `DifficultyCalculator` not implement yet, so change test case actual value is ok.

PR #551, opened 2021-04-10
Enable to set max and min available font size in bindable.

andy840119:
It's weird to let user set some text like main lyric tooooo small.
So should have a range of text size to let user able to adjust.


andy840119:
Not good at typing at morning 😢

Issue #530, opened 2021-03-28
Implement select font from cached


andy840119:
What should be done in this pr:
- get all font name in cache folder.
  - Not really sure should have manager to handle that?
  - Maybe called `FontStorage: LegacyDirectoryArchiveReader` or something (not `FontStore`, there are different class)?
  - [ ] should make a check that what's different between `DirectoryArchiveReader`  and `Stoarge`? which is better to use.
- Have an option to customize font and selected font.
- Have notification if no fonts in there.
- Have folder to open the folder to let user place the font.
  - Create default font and cache folder before click the folder.
  - Not really sure should place a readme text file into there.

andy840119:
Also notice think how to register  preview font store logic.
And might need to un-register font if change preview font. (maybe destroy then re-create again?)
.
Or maybe rename from `KaraokePlayfieldAdjustmentContainer` to FontRegisterLayer ?
Or check now to create karaoke ruleset's own `SkinProvidingContainer`?
Or use `ArchiveModelManager`, like how `SkinManager` do?

andy840119:
Maybe should create something like `FontInfo : IHasFiles<FontFileInfo>`

Also if font info has been saved, what should be saved?
- id
- name
- family
- weight
- files(real file path)
- equal
- available chars(for fasting checking if not in drawable), not sure how to deal because it might be large.
- support language(?)

andy840119:
Ask a simple question:
is font works like skin?
.
And will have something like this?
![image](https://user-images.githubusercontent.com/9100368/114259128-aa258900-9a06-11eb-86fd-71a937416134.png)

andy840119:
Implemented.

Issue #520, opened 2021-03-22
Font selector in karaoke ruleset.


andy840119:
Unlike other ruleset, this ruleset might have it's own font store.
Prepare lots of fonts for karaoke songs.
.
![image](https://user-images.githubusercontent.com/9100368/112019281-a07cd400-8b72-11eb-9377-a2db2b2e083b.png)
So might need some dialog to select and preview fonts.
Font should return `FontUsage` struct and can be save / load (might with json format)
.
Also font might be able to group by language.
Or adding tag(like this song is for Japanese songs) for fast search.
.
And here's the property in `FontUsage` :
- FontName(string) -> combine with `Family` and `Italics`
- Family(string)
- Weight(string)
- Size(float)
- Italics(boolean)
- FixedWidth(boolean)

andy840119:
Think about the style.

andy840119:
Also think about how to store the font (class name)
Font should have those option:
- Name
- English name
- Family(use to find real font.)
- Recommend localization.
- List of Weight(Maybe include `Italic`)?

Note : only `OpenSans` font in `osu-framework` use `Italics` property.

andy840119:
This info can might be able to stored into dialog now.
But should find a time to move this shit out.

andy840119:
Implemented.

PR #552, opened 2021-04-10
Implement font manager.

andy840119:
This is the implementation of #520 and #530.
What's changed in this PR:
- Implement font manager, which can provide
  - Get the list of default fonts
  - Get the list of imported font
- Inject font manager to game host(might think about how to remove cache in future)
- Fix import font logic
- Apply font if has the default value.


andy840119:
Waiting for CI happy.

andy840119:
Ass

Issue #585, opened 2021-04-30
Refactor IPositionCalculator


andy840119:
For now, the column width is fixed.
should be able to change by the skin.
.
Maybe see how mania ruleset works.

andy840119:
todo : 
get skin column height in position calculator.
also should think about how to update note position if skin column height changed?

andy840119:
Also, need to think about why need to use `IPositionCalculator`?
Got no reason why need to do that 🤔 
.
or why not just combine with `scrolling note playfield`?

andy840119:
maybe position calculator should be able to get value from `ISkinSource`.
and another question is how to let note able update position if column height changed?
maybe just use bindable 🤔 ?

if use bindable also, should place bindable in every note or just in note playfield?
maybe just follow how scrolling info does?

andy840119:
Maybe called something like `IXXXInfo` ?

andy840119:
Maybe remove position calculator into another class (maybe something like position algorithm?)
Also should write some test case.

andy840119:
And this class should expose:
- Saiten tone offset adjustment.
- column height and column spacing (maybe should cancel column spacing on one day?)

andy840119:
maybe note position should be auto-calculated in `ScrollingHitObjectContainer`
But `CreateHitObjectContainer()` is marked as sealed in `ScrollingPlayfield`

andy840119:
See `LegacyKaraokeSkinConfigurationLookups` to get stage height and spacing.

andy840119:
Close due to have solution already.

Issue #575, opened 2021-04-24
Able to compare multi edit note playfield in lyric maker.


andy840119:
This is the part of the answer for issue #569 
.
Pros of the idea:
- Better to manage note.
- Able to check if lyric has its note object, and should give it an alert if no object in here.

andy840119:
This issue should be consider : 
What if the case user wants to see(for comparing) other lyric notes in lyric editor?

andy840119:
Maybe create a copy button, then will open a new dialog and select matched lyric?

PR #587, opened 2021-05-01
Sync skin component

andy840119:
What's changed in this PR:
- move some class into the `legacy` namespace.
- move some class from `piece` into `legacy` namespace, like `default body piece`.
- should able to get column height from karaoke skin, and apply in note playfield and note.
- move judgement area percentage into the skin.
- adjust preview skin for lyric maker.


andy840119:
Also, this issue is related to 
- part of adjusting for #575 
- most implement part for #585

PR #589, opened 2021-05-02
Update NuGet 

andy840119:
Update NuGet and fix compile error.


andy840119:
Wait
No error?
🤔

andy840119:
Tested and it works in the newer version of vs

PR #591, opened 2021-05-03
Implement default lyric piece

andy840119:
Separate karaoke lyric in drawable lyric into a new piece, called `DefaultLyricPiece`
.
Why this change needed:
- `DefaultLyricPiece` is focused on text and style display, remain part like layout will be handled in the parent.
- some case not wants to use the whole feature of drawable lyrics, so `DefaultLyricPiece` might be a better choice for use.
- fix style broken in the lyric editor.
- better for future works(V2).


andy840119:
Test it and it works in both editor and gameplay.
This change is a good direction and might be able to merge it now.

PR #600, opened 2021-05-09
Use lazy way to fix CD not working

andy840119:
because ILRepack not working in .net5
so should roll-back environment from .net5 to .netCore 3.1

Caused in #413


andy840119:
Got no idea will appear will use .netCore 3.1 or net5 🤔

andy840119:
oops shit

andy840119:
https://github.com/actions/setup-dotnet
See this article on how to assign a target net core version.

PR #605, opened 2021-05-11
Refactor note selection handler logic.

andy840119:
todo : might think about using which is better.
composer logic or customize selection blueprint.


andy840119:
Can be closed due to implemented.

Issue #606, opened 2021-05-12
Able to drag to select time-tag in time-tag edit mode.


andy840119:
User should be able to drag and select multi time-tag in time-tag edit mode.
Need to think about:
- is this feature really needed because already contains time-tag editor in overlay.
- if need, should sync selection with time-tag editor in overlay.
- will conflict with current cursor algorithm?

andy840119:
This priority is lower than #601

andy840119:
Wait
Need to think about is this feature really useful?
It might not really need that unless the user wants to:
- drag and delete
- drag and move(?)
- drag and create(not useful at all)

andy840119:
Also, is that a good decision to make all getXXXPositionGetter into edit lyric piece?

andy840119:
Update:
time-tag is not draggable now.
it's for previewing overlay selection.

andy840119:
And because this blueprint is not supported draggable now.
This issue can be closed.

PR #612, opened 2021-05-17
Implement time tag blueprint

andy840119:
This is the first implementation of #606 
Some of the features still need to be considered.
Like should able user able to select time-tag be directly dragging because it'll cause select time-tag feature broken...
.
What's changed in this PR:
- [x] implement `TimeTagSelectionBlueprint` and `TimeTagBlueprintContainer`
- [x] move some of getting position method to `EditorLyricPiece`
- [x] refactor get hover position function.


andy840119:
Fuck that codefactor

PR #627, opened 2021-05-22
Update nuget package

andy840119:
Update NuGet package and fix some compile error.

todo : might test in PC?


andy840119:
lazy to do that.
just give it a pass if CI says ok.

Issue #601, opened 2021-05-10
Able to edit ruby / romaji tag.


andy840119:
Because now blueprint is not restrict to hit object only.
So it,s able to implement that

andy840119:
Implement step(notice that this step only works in ruby and romaji, main text might not so important in this pr):
1. create  TextTagBlueprintContainer : BlueprintContainer<TextTag>
2. implement TextTagHitObjectBlueprint : SelectionBlueprint<TextTag>

it might support create blueprint with class only, not need to create real drawable object.

also need to think about is this need to add "Editor" text.

andy840119:
And those behavior should be supported also:
1. should be able to drag position by single / multi selection.
2. should be able to edit text by single / multi selection by double click, and new modal component is needed.
3. should be able to delete by single / multi selection by right click.
4. should be able to add new ruby/romaji, need to think about how to implement that.

Also, should have:
- create(if not select exist time-tag)
- edit
- delete

by right-click.

andy840119:
Also, should consider:
1. make ITextTag as base class because got no idea why make it as interface before 🤔 
2. might make start and end index in text tag has bindable also(for dragging event in blueprint)?
3. make text as bindable (this can be implemented later.)

andy840119:
And might have background in ruby and romaji drag area to let user know only ruby and romaji is draggable.
see how [time-tag editor](https://github.com/karaoke-dev/karaoke/tree/master/osu.Game.Rulesets.Karaoke/Edit/Lyrics/Overlays/Components/TimeTagEditor) works

andy840119:
Remain : 
- [ ] Click create button to crate ruby/romaji tag at dragging position, should see how placement works. (Note: placement blueprint still needs HitObject)
- [x] Drag to change ruby/romaji position
- [x] Able to change start/end index.
- [x] Able to double click to change the text.

andy840119:
The idea from #621
Maybe it's able to have right-side spacing to place a list of ruby/romaji from selected lyric.

andy840119:
Can be closed due to remaining parts (create ruby/romaji) will handle in #618

PR #635, opened 2021-05-25
Able to delete ruby/romaji tag.

andy840119:
Implementation of #601


andy840119:
Seems code factor is absent

PR #647, opened 2021-05-29
Implement lyric preview.

andy840119:
image
What's changed in this pr:
- implement lyric preview
- works in translate section
- works in ruby/romaji display section.


andy840119:
Remain todo is about applying preview change if the use changes the lyric text.
Already open a new issue to trace that.

PR #663, opened 2021-06-03
Use more generic hit object composer instead of origin one.

andy840119:
In before, this part inherits HitObjectComposer and it might cause some problems in the future.
e.g: 
- might have API broken if HitObjectComposer changed.
- hard to maintain.

So should write its own class for that, even has duplicated logic.


andy840119:
Some of the logic might be broken but will fix later.

Issue #693, opened 2021-06-14
Add right-side extend in edit note mode.


andy840119:
Might have those sections:
- auto-generate saiten graph by voice file(for placing note)
- should be able to adjust generate the result. 
- auto-generate note position(should finish the first step) -> need to think a better way on how to apply in target lyric.
- show list of invalid note result

All the details will be listed in notion:
https://www.notion.so/auto-generate-note-position-by-time-tag-038fd799c10c4666a67bc3c3d60a74bb

PR #697, opened 2021-06-18
Add right-side extend in edit note mode

andy840119:
First, implement #693 

What's added in this pr:
- implement note extra area.
- fix some issues.

Note: auto-generate saiten path will be moved into another pr.


andy840119:
image
The Pitch result (from singer voice only) is messy, not really sure is because the voice is from a woman.
The result is playback with 0.75x and a little better(?)


https://www.speechandhearing.net/laboratory/ampitch/
It tests smoothly if the voice is from man.

PR #702, opened 2021-06-20
Implement note generator.

andy840119:
Implement a base feature note generator.
Like other generators, will have config.
Input lyric and will got some notes.


andy840119:
Although this generator is very simple, just let it pass now.

Issue #448, opened 2021-02-01
Able to select lyric for apply in lyric editor.


andy840119:
Seems there will be more change to access batch change property in list of lyric (like #443 to apply generated result to lyric or #410 to lock lyric)
Might should have dialog to select batch of lyric.
And this component will be placed into `Edit/Components`, named `LyricSelectionDialog`.

andy840119:
Or just select in lyric editor?
And apply select to preview.

andy840119:
Got a fucking UX for selection lyrics to apply.
Will be used in the auto-generate part first.
.
Will have a checkbox in front of each lyric line.
So might not need dialog.
.
Preview might not that easy to be implemented if made this feature in lyric editor 😢

andy840119:
Implemented.

PR #703, opened 2021-06-20
Implement auto generate note section

andy840119:
Implement auto-generate note section.

what's add in this pr:
- split note edit mode number from 1 into 3.
- add a section to let the user able to switch.
- implement note auto-generate section.
- add the auto-generate note feature in the note manager.

todo : 
- [ ] should update the note field of creating or delete notes.
- [ ] should be able to select target notes (need to wait until #448 implemented).


andy840119:
Will have other issue to handle todos.

Issue #696, opened 2021-06-18
Rewrite note selection blueprint to use only HitObject (no DrawableHitObject) for layout


andy840119:
See how this PR implemented: 
https://github.com/ppy/osu/pull/13499/files

andy840119:
Fixed.

Issue #582, opened 2021-04-30
Fix drag note position in lyric editor edit note mode.


andy840119:
Open issue #580

andy840119:
Got the reason:
`moveEvent.ScreenSpacePosition` not changing inside playfield
got no idea about that.

andy840119:
Also should handle breaking change caused in #663 also.

andy840119:
Fixed.

PR #771, opened 2021-07-20
Fix note drag issue in editor

andy840119:
image
image
Close code quality issue #696
And fix note dragging is weird in issue #582


andy840119:
also should notice that this pr does not contain dragging notes in the lyric editor.

this issue is not hard to fix.
but need to come out with a better way to collect the same logic.

Issue #774, opened 2021-07-22
Should sync selection in singer container.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/126644282-ebbc45d6-a133-484b-95a4-d701b79b974e.png)
It might be weird to let this shit happened.

should sync all of them.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/126644548-a057ee65-0ce8-4e3b-8a72-464671344458.png)
Also should fix this case if a single disabled lyric is selectable, but cannot be multi-selected.

andy840119:
And should clear selection after assign lyric to target singer.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/126645030-a7dfaa64-df84-44b8-acd5-f2b114f0e629.png)
#691 still not fixed.

andy840119:
Check those issues.
Should be fixed.

andy840119:
Implemented.

Issue #691, opened 2021-06-13
Fix lyric singer list should roll-back to default if singer deleted.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/121799854-6ac72300-cc69-11eb-80be-7ba7823f85b5.png)

What's should be fixed:
- [x] position is hard to sync if zoom out to min value.
- [x] should roll back singer in the lyric is singer deleted.
- [x] scale and position are different if create a new singer.
- [x] should update right-side selection if singer deleted.

what's can be improved:
- [ ] should show the edit singer dialog once click add singer button.

andy840119:
Also should have base implement avatar section, might have something like `LabelledImageSelection` maybe just inherit `FileChooserLabelledTextBox`?

andy840119:
And should have something like `LabelledColourSelection` to select the singer color.
No need to implement it because will have an official one eventually.

andy840119:
Fixed.

PR #775, opened 2021-07-22
Should sync selection in singer container.

andy840119:
Closes issue #774 and #691.
And refactor some code.


andy840119:
Tested.
Seems nice.

Issue #778, opened 2021-07-24
Should have preview font style in font selector.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/126869870-708f382e-dd02-4536-bea9-c490d3b86fed.png)
Do not have makinas4 font style.

.
![image](https://user-images.githubusercontent.com/9100368/126869902-7f6b5aab-5afc-4b70-b793-5f99f94c6e89.png)
And this is what this font should look like.

andy840119:
Also, the right side should have preview also.

andy840119:
maybe should write utils to convert `FontUsage` into `FontInfo`.
And should create `GlyphStore` by `FontInfo`

andy840119:
And need to think of a better solution in:
- gameplay
- font dialog
- preview in the config page.
- might be in other places like the editor.

And should notice that:
- should not load all the fonts.
- should be able to hot-load new font 
- should be able to clear font if not needed.

Maybe should add extensions for `FontStore`:
- able to add store by `FontInfo`
- able to remove store by `FontInfo` (optional, need to check able to implement that)
- able to clear store.

andy840119:
Implemented.

Issue #772, opened 2021-07-21
Fix lyric font broken if playback.


andy840119:
the reason is caused by not apply the font because it only initialize in `ApplySkin`.
but `DrawableLyric` might use several times.
.
And not really sure is this the reason that makes the second lyric broken.

andy840119:
`DefaultBodyPiece` might have same issue.

andy840119:
Check if fixed in #1005 ?

andy840119:
Seems fixed.
tested in osx only.

PR #779, opened 2021-07-25
Should have preview font style in font selector.

andy840119:
Implement part of the issue in #778.

Config dialog can preview customize font.
But config preview section should still need to implement.


andy840119:
related to #772 
need to refactor the skin part.

Issue #791, opened 2021-07-30
Implement for supporting otf text format.


andy840119:
Implement something like OtfGlyphStore / TTFGlyphStore to let karaoke ruleset support oft / ttf format easily.
And might be able to use this repo:
https://github.com/Robmaister/SharpFont/tree/master/Source/Examples

Notice that it's just a easy to let user import it's own  oft / ttf font.
But not means it can be distribute unless it's able to compress oft / ttf with target charset.

update:
cannot open font from the stream
氣死

andy840119:
Inspired from #787 and [this chinses article](https://www.twblogs.net/a/5b8081682b71772165a8011e).
And looks like .fnt is not the only choice for osu-framework when ppy design it.

andy840119:
Seems have another choice:
https://github.com/LayoutFarm/Typography

update:
The demo is crash-friendly.

andy840119:
Or the safty way is using sixlabor
https://github.com/SixLabors/Fonts

update:
Fick this shit does not implement its one text render.
And most of demo is outdated
WTF

andy840119:
another repo:
https://github.com/vidstige/NRasterizer
not upload to nuget 😢

andy840119:
https://www.grapecity.com/documents-api-imaging/pricing
paid solution

andy840119:
And another issue:
otf/fft is vertix format.
And it can be render in openGL directly.
Not really sure if weird to change to image first 🤔

andy840119:
Seems Sixlabor is not supported `.otf` now.
Also should implement another font store after mark sure that TTFGlyphStore is workable(without cost toooo much time to generate texture.)

andy840119:
I guess it might be better to wait until Sixlabor supports of instead of finding a new package.

PR #794, opened 2021-08-01
Implement TTFGlyphStore

andy840119:
This is the part of the implementation of #791 
but sadly, it does not support `FontStore` now.
should think of a better solution, or wait until official to support that 🤔


andy840119:
Not really sure this shit is cross-platform?

andy840119:
tested and it's able to export char image in osx.
so it's ok to give it a merge 👌🏼

PR #800, opened 2021-08-08
Use new switch expression.

andy840119:
See https://docs.microsoft.com/zh-tw/dotnet/csharp/language-reference/operators/switch-expression
I think it's able to use it now because in some cases it will reduce lots of lines.

update:
it will save almost 50% of lines if change to switch expression.


andy840119:
image
Master-999

PR #803, opened 2021-08-14
Fix change log style

andy840119:
After:
image

What's fixed in this PR:
- Fix the second line will at the top of the first line.
- Remove some duplicated config.
- Most components should inherit osu's one.


andy840119:
image
Also got a issue that if inherit `CreateSpriteText` in `ChangeLogMarkdownTextFlowContainer` can see:
Although will got `OsuSpriteText` in `OsuMarkdownContainer`, but might get the different font.
Got no idea why IL code in nuget is different from the source code on the repo.

andy840119:
Change log font issue can be later.
Because wiki in Lazer has the same things.

Issue #804, opened 2021-08-14
Font selection improvement


andy840119:
![image](https://user-images.githubusercontent.com/9100368/129446315-42830af3-4bd6-4b79-9c71-90f258f5bc60.png)

Now it's able to change the font without any conversion 🥳 
The next step is UI improvement.

The conversion will include:
- [ ] Check why ruby/romaji cannot apply the customized font.
- [x] Should mark font extension at the end of the font (if the font is not internal)
- [ ] [optional] Should load font list only if user opens the select font dialog.
- [ ] [optional] should limit max font number in the dictionary (for the case if user load too many fonts)

andy840119:
Also:
- [x] should have the ability to hot-reload the font list if the dictionary is changed in the font selector.
- [x] should be able to open the dictionary on the config page.
- [x] should create a folder at default. Need to check should do this job in `FontManager` or another place.
- [ ] [optional]should be able to hide internal fonts.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/132245901-ed2c0a5c-6d03-4f64-ba6c-c518d7d170ce.png)

About font missing issue.
I guess it's about getting the wrong base height(getTextHeight) in PositionTextBuilder

`baseHeight.HasValue` should be less then 1 but got no idea why got 13 in customized font.

andy840119:
Font selection is useable state.
Also, it's already able to mark the font type.

PR #806, opened 2021-08-14
Add font type property

andy840119:
It's a prepare for showing font format in font selection listed in #804

What's implemented in this PR:
- `FontInfo` should have format property.
- if wants to get `glyph store` from `font manager`, will see `font format` instead of search all folder.
- `KaraokeLocalFontStore` now can accept add font by font usage, which will cost less effect on changing format after.


andy840119:
If need to display font family with font format, should:
- should add overridable function in `RearrangeableTextListContainer` to let it able to set `DrawableTextListItem`'s text, include drawable.
- `familyProperty` in `FontSelectionDialog` should be `TextPropertyList<FontInfo>`, or just make something like `FontFamilyListContainer<T>` with list of fonts in ctor.

PR #810, opened 2021-08-20
Fix TTF font cannot use in release.

andy840119:
Fix issue #809

PR #812, opened 2021-08-20
Fix TTF font cannot use in release.

andy840119:
Fix wrong wrapping dlls caused in #810


andy840119:
Lazy to run CI, just give it a pass.

Issue #837, opened 2021-09-05
Make a check that why some section page is disappear


andy840119:
![image](https://user-images.githubusercontent.com/9100368/132132067-caa2c35d-0a67-44ba-85c5-fcd24d1e7f40.png)
Check this shit.

andy840119:
Seems it's not showing?
Not really sure does base class has been hidden in default?

andy840119:
Can close now because found the root cause.

PR #838, opened 2021-09-06
Fix the case that some section page is disappear.

andy840119:
Use temp way to fix this shit found in #837
might have compile error in next release.
See the change in https://github.com/ppy/osu/pull/14554/files


andy840119:
Let this shit merge and fix it in the next release.

PR #852, opened 2021-09-21
Able to adjust ruby romaji margin in config page.

andy840119:
What's added in this pr:
- Able to adjust ruby romaji margin in the config page.
- Fix translate is disappear in the lyric if the translated language is not Japanese.
- Fix not hide ruby/romaji if user select to hide them.


andy840119:
Also should hide ruby/romaji/translate config if the user hides them?

Issue #854, opened 2021-09-22
Implement local shader things.


andy840119:
For letting user able to load their customized shader.

How:
- Create `LocalShaderManager` inherit `ShaderManager`, which is a `ResourceStore<byte[]>` to provide a list of the local shaders.
- Make a check that other components like `PostProcessingContainer` can access `LocalShaderManager` without any BDL issues.

andy840119:
Note: maybe can use `LocalShaderManager` to get list of shaders, then place it into something like `KaraokeSpriteText`

andy840119:
Notice that local shader manager will not be added until able to let users able to customized their shader.

PR #856, opened 2021-09-30
Update font framework

andy840119:
What's changed in this PR:
- update font package.
- apply shader instead of property in karaoke font.

Also implement issue #854


andy840119:
Notice that local shader manager will not be added until able to let users able to customized their shader.

andy840119:
tested and shader works on release version.

Issue #862, opened 2021-10-11
Should move SettingHUDOverlay into HUDOverlay?


andy840119:
Because seems `drawableRuleset.Overlays` is placing drawable for gameplay.
Maybe it will be better if move it into HUDOverlay?
.
And notice that `AllowGameplayOverlays = false` will only affect some overlay.

andy840119:
Might make something like:
1. Make setting button area inherit `ISkinnableDrawable`m see `PerformancePointsCounter`
2. See `CatchLegacySkinTransformer` how to place customized setting button.
3. Call `SkinnableTargetContainer` to place setting button area.
4. `setting button area` should be able to get `HUDOverlay` and place the setting overlay into here.

After that setting overlay button should be able to adjust button position.
And that button should be placed into `UI.HUD`

andy840119:
Oops. forgot that how to get `KaraokeSessionStatics` in `HUDOverlay` area 🤔

andy840119:
Implemented.

Issue #859, opened 2021-10-06
Check why time scrollbar is disappear in auto-play.


andy840119:
IDK why other ruleset still has the scrollbar, maybe mod API changed?

andy840119:
![image](https://user-images.githubusercontent.com/9100368/136776155-dcf61d27-c8e0-4349-83d9-c957bff1d378.png)
Wait,mode icon not visible also.
Seems HUD has been hiddened.

andy840119:
Guess it's caused by this line:
```
public override bool AllowGameplayOverlays => Beatmap.IsScorable() && !Mods.OfType<KaraokeModPractice>().Any();
```
in `DrawableKaraokeRuleset`

andy840119:
After #863 Merged, should have a checkbox in overlay to decide should hide all components in HUD overlay if beatmap is not scorable?

PR #863, opened 2021-10-16
Move SettingHUDOverlay into HUDOverlay

andy840119:
Implement issue #862 
for preparing to solve #859


andy840119:
Just ignore code factor.

PR #875, opened 2021-10-28
Use `Storage.PresentFileExternally` when opening lyric files after export

Susko3:
Uses the newly-added `PresentFileExternally` to show files in windows explorer.
This has the added benefit of selecting/highlighting the file, making it easier to know which file was just exported.
(The selecting/highlighting part currently only works on Windows.)

I made the upstream change in https://github.com/ppy/osu-framework/pull/4831, so I wanted to get consumers updated to the new feature.
That's why I'm making this PR.

### DIFF
--- a/osu.Game.Rulesets.Karaoke/Edit/Export/ExportLyricManager.cs
+++ b/osu.Game.Rulesets.Karaoke/Edit/Export/ExportLyricManager.cs
@@ -23,8 +23,9 @@ namespace osu.Game.Rulesets.Karaoke.Edit.Export
         public void ExportToLrc()
         {
             var exportStorage = storage.GetStorageForDirectory("lrc");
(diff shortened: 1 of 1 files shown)


andy840119:
I'm not sure does `PresentExternally` means open the folder.
And `PresentFileExternally` means open the folder and focus on the target file.
Is that right?

Susko3:
Yes, that's right.
We also have `OpenFileExternally`, but your use case is more suited for `PresentFileExternally`.

https://github.com/ppy/osu-framework/blob/0ec0502f7ddd01785afcf52266fb4436722545e3/osu.Framework/Platform/Storage.cs#L110-L128

andy840119:
OK.
Seems `PresentFileExternally` might be the better behavior.
Thanks for your contribution.

PR #885, opened 2021-10-31
Update font package

andy840119:
Update font package and fix compile error.


andy840119:
And link that the main change is about improving outline shader performance.
And it's fixed by https://github.com/karaoke-dev/osu-framework-font/pull/78

PR #905, opened 2021-11-09
Refactor style part in the skin system.

andy840119:
Will be merged until almost implemented.


andy840119:
I guess it should be OK to make a merge after testing on windows.
Layout and config part can do it after.

PR #939, opened 2021-11-26
Fix missing dependencies not found issue in karaoke editor.

andy840119:
Fix some dependencies not being cached.


andy840119:
Got no idea why another two bots is not here 🤔

Issue #930, opened 2021-11-24
Implement new screen for karaoke skin editor.


andy840119:
Seems that the editor might be able to navigate to the new screen, so it will be able to edit the skin in the new screen.
And this editor will be used in:
- editor
- config page. 

So should think about:
- should be better to place it into the screen namespace? or just place it into the edit namespace?
- naming
- ctor, might place something like `KaraokeSkin`?

andy840119:
Should be able to switch to every tab, like how #937 do.

PR #942, opened 2021-11-27
Implement karaoke skin editor

andy840119:
This is the first implementation of karaoke skin editor.
Might still have bugs, but just let it merge if it can pass the CI.


andy840119:
Base implementation of #930

Issue #889, opened 2021-10-31
Should implement page for auto-generate config?


andy840119:
![image](https://user-images.githubusercontent.com/9100368/139592280-13d47ec7-4780-42ae-b52c-e07242018919.png)
It might be easy to edit if change style like that.

andy840119:
And this can be remained until support full-screen dialog in osu after implement.
.
[!!!] Also should test that is BDL works or not, it might be not working.

andy840119:
Note:
- Might have different config setting if use different lauguage for time-tag generator.

andy840119:
Implemented.
But not really sure should have a page for all generator config.

PR #989, opened 2021-12-26
Implement auto-generate config popover in the lyric editor page.

andy840119:
Should be able to adjust  auto-generate config in editor page now.


andy840119:
This is the implementation of #889.

Issue #1041, opened 2022-01-16
Should record HUE instead of Hex in the singer.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/149654896-1d64a369-ca06-4e82-b70a-8a31e831e694.png)

As #1040
It might be an eye pain if record hex code.
should record hue value instead.

Pros:
- prevent users select the dirty color.

Cons:
- can't select black or white as a singer color.
- not much choice for users.(maybe it's not a bad thing, i guess.)

andy840119:
Need to get a hue color picket and lock brightness and darkness property.

andy840119:
This issue can be closed until hue color picker has been implemented.

andy840119:
Implemented in the #1224

PR #1042, opened 2022-01-17
Should record HUE instead of Hex in the singer.

andy840119:
Implementation of #1041

Before:
image

After:
image


andy840119:
Before:
image

After:
image

PR #1071, opened 2022-01-26
Update osu.game to the latest.

andy840119:
image
Fix lots of breaking change.
Still have test cases failed issues to be fixed 😢 .
.
What's still need to check by a human after all test cases passed:
- [ ] can read the beatmap in gameplay. -> cannot get any customized beatmap now.
- [x] ruleset config is OK in the test cases.
- [x] fresh light mod. -> working but weird.
- [x] should make sure still able to access the editor. -> not able to access it now. -> fixed.
- [ ] check if skin works normally
- [x] check if create `KaraokeBeatmap` missing ruleset info.(might cause crash.)

皮皮你一定要這樣嗎?


andy840119:
https://github.com/ppy/osu/pull/16632/files
Song selection might be fixed after new version update.

andy840119:
the editor is still broken.
seems not possible to be fixed right now.

andy840119:
Maybe should create another PR about how to generate a customized beatmap.
But seems it's not affected development now, so give it a merge.

Issue #1073, opened 2022-01-28
Apply new skin structure.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/151581371-046594b7-0de7-461e-9499-220e15cdbc3a.png)
https://miro.com/app/board/uXjVOdFaTlU=/?invite_link_id=383662171171
This is what the new skin structure looks like.
And seems there's no issue for this structure to replace the current one(even with the V2 one)
So can be changed without any concern(maybe).

andy840119:
The last steps before implementation:
- [x] should think about how to assign the layouts to the lyric. -> right-click to the lyric, will see the mapping names in the selection.
- [x] how to manage layout property in the list of lyrics. -> will have a page to show all mappings that apply to the lyric.

andy840119:
Will be closed until the editor has been implemented.
I think it should be a big topic so should continuously track it until release.

andy840119:
This issue can be closed because we use stage to handler the mapping role instead.

PR #1076, opened 2022-01-30
Apply new skin structure into karaoke beatmap

andy840119:
Part of implementation of  #1073
This PR is focused on changing the structure in the skin and beatmap skin.


andy840119:
This issue might have a breaking change:
- because remove list of bindable in the skin, so change size in editor might not working.

PR #1083, opened 2022-02-03
Apply into legacy skin conversion

andy840119:
After skin format changed in #1073
Should be able to get the mapping roles from the legacy beatmap skin.
.
todo:
- [x] should test if works in the gameplay.


andy840119:
seems todo list bot is busy.

Issue #1081, opened 2022-02-03
Have ApplyTo() function in the skin element.


andy840119:
Inspired from #1073
.
Because we might have more and more skin element types.
So it might be a good idea to move all the change style/layout logic into skin element.
.
e.g:
```csharp
public class LyricStyle : IKaraokeSkinElement
{
    public void ApplyTo(Drawable d)
    {
        if (d is not DrawableLyric drawableLyric)
            throw new InvalidDrawableTypeException(nameof(d));

        foreach (var lyricPiece in drawableLyric.lyricPieces)
        {
            // Apply shader.
            lyricPiece.LeftLyricTextShaders = SkinConvertorTool.ConvertLeftSideShader(shaderManager, this);
            lyricPiece.RightLyricTextShaders = SkinConvertorTool.ConvertRightSideShader(shaderManager, this);
        }
    }
}
```
.
But it will have some problems:
- not easy to get the DI -> resolved
- might have several parts needed to apply the element value. e.g: drawable note and default body piece need the same config. -> maybe should be able to expose the skinnable drawable?
- might access the private properties. -> technically should make every needed property be public.

andy840119:
reference:
- `ApplyTransforms()` in `StoryboardSprite`
- `ApplyToDrawableHitObject()` in `IApplicableToDrawableHitObject`

andy840119:
Maybe:
```
public void ApplyTo(IReadOnlyDependencyContainer dependency, Drawable d)
{
    var skinManager = dependency.Get<SkinManager>();
}
```

andy840119:
Another benefit for this change:
- we can make something like `LyricAnimationScript: IKaraokeSkinElement` for able to write a script to control the lyric or other object.

andy840119:
Implemented.

PR #1084, opened 2022-02-03
Have ApplyTo() function in the skin element

andy840119:
Trying to resolve issue #1081


andy840119:
I think it should be the right direction in this change
- [x] test in the gameplay. -> tested in osx only lol

Issue #1079, opened 2022-02-01
Steps to remove all layout index in the lyric


andy840119:
It will be a important step for implementation of #1073
.
Will need to be resolved:
- [x] remove layout section from lyric editor.
- [x] deal with layout generator(maybe should be moved into `KaraokeLegacySkinTransformer`, maybe used to create fake layout group?)
- [ ] deal with `KaraokeBeatmapProcessor` because layout index might affect appear time in the lyric(technically it should be calculated in the editor, not in the processor).

andy840119:
Might write a Generator for calculating the lyric start/end time by time-tag. 
-> config: number of lyrics that in the screen.
-> think about how to move this feature into the lyric editor.

andy840119:
Seems can close this issue.

PR #1093, opened 2022-02-05
Remove layout index from lyric.

andy840119:
It's the last step of #1079
What's done in this PR:
- remove copy layout Id logic in lyric utils.
- should not able to assign layout index directly in the editor.
- should trigger the skin change instead if re-assign the layout id. -> it's a todo issue
- remove broken test case.
- should fix the preview layout in the lyric -> it's a todo issue
- finally can remove this property.


andy840119:
All relative bugs are opened.

Issue #1099, opened 2022-02-05
Seems internal skin is not loaded well in release build.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/152643737-8bd86c01-ed7e-49bc-9952-40f506839aa5.png)
Might be caused due to the IRMerge.

andy840119:
Step:
1. should not use `GetType().Assembly` in the `InternalSkinStorageResourceProvider`
2. `GetAssemblyByName(string name)` in `AssemblyUtils` should compare the full name because `osu.Game.Ruleset.Karaoke` might get the result `osu.Game.Ruleset.Karaoke.Tests`

andy840119:
Fixed.

PR #1100, opened 2022-02-05
Seems internal skin is not loaded well in release build.

andy840119:
trying to fix issue #1099.


andy840119:
image
Worked in release build, nice.

Issue #1101, opened 2022-02-05
Should be able to show the release version someplace.


andy840119:
Should be able to show the release version someplace(e.g: the config page).
It might not easy to make sure if using new dll without version info.

andy840119:
There are two steps that need to be implemented:
- should change the version by tag name.
- should get the version from main dll.

andy840119:
Check how `AssemblyVersion` in `OsuGameBase` do.
```csharp
var version = Assembly.GetEntryAssembly()?.GetName().Version ?? new Version();
var versionText = $@"{version.Major}.{version.Minor}.{version.Build}-karaoke";
```

andy840119:
![image](https://user-images.githubusercontent.com/9100368/153432589-6af25f1e-5233-4ab7-95b8-9575ceb632b9.png)
Already contains tag in release version lol

andy840119:
Implemented.

PR #1112, opened 2022-02-10
Should be able to show the release version someplace.

andy840119:
Implement issue #1101
image


andy840119:
image
Seems it works in the release build also.

Issue #1113, opened 2022-02-12
Some improvement about time-tag mode in the lyric editor.


andy840119:
What can be improved:
- [x] Should be able to record the zoom scale in record/adjust mode.
- [x] Should be able to show/hide the waveform graph in the record/adjust mode.
- [x] Should change editor clock time if change caret position in record mode.
- [ ] Should have tap area in editor mode because it might be hard for the user to remember the hot-key for setting the time.

andy840119:
Guess should be able to closed due to most parts has been done.

PR #1114, opened 2022-02-12
Create mode-specific state for lyric editor.

andy840119:
Changing lyric editor state for pre-step of issue #1113 
It will be better to make mode-specific state classes for able to add more state properties with different edit modes.
.
https://miro.com/app/board/uXjVOdFaTlU=/?invite_link_id=156545363905
Before:
image
After:
image
.


andy840119:
Create an issue for tracking is there any possibility to improve the code.
But this PR is good enough to be merged.

PR #1124, opened 2022-02-16
Update nuget package

andy840119:
What's done in this PR:
- upgrade some package to the latest.
- move the target of .net to 6.0.

PR #1128, opened 2022-02-21
[NOT WORKING] Trying to let ILRepack.MSBuild.Task works again.

andy840119:
ILRepack.MSBuild.Task not working after #1124.
And this error is caused because upgrade `Microsoft.CodeAnalysis.NetAnalyzers` into 6.0.0
.
so remove un-needed package because it will cause compile error in .netCore 3.1
and ILRepack.MSBuild.Task can only run on this environment.


andy840119:
image
still not working 😢

Issue #1107, opened 2022-02-07
Implement karaoke beatmap JSON loader.


andy840119:
Guess it's time to migrate the temp beatmap to JSON.
The first step might be able to read the customized karaoke JSON beatmap. 
.
the format might be:
```
// karoake beatmap v1
{
  ...json string
}
```
And will load the beatmap only if start with `// karaoke beatmap v1`

andy840119:
Guess the most important part in this issue is able to read the json beatmap, so just ignore the reference mistake issue.
.
Also, it will be great if can export karaoke beatmap with json format(for testing purpose).

andy840119:
Just make a screenshot code:
```csharp
// because it will cause invalid stored block lengths issue
// so create new zip file instead of open then add the file.
// https://github.com/adamhathcock/sharpcompress/issues/269
string str = item.GetDisplayString().GetValidArchiveContentFilename() + FileExtension;
using var stream = exportStorage.GetStream(str, FileAccess.ReadWrite, FileMode.Create);
using var jsonBeatmapStream = getJsonBeatmapStream();
using var archive = ZipArchive.Create();
archive.AddEntry(filename, jsonBeatmapStream);
archive.SaveTo(stream);
```
got no idea why this shit can only add 0 bytes of file in the zip.

andy840119:
Guess implemented.

PR #1130, opened 2022-02-23
Implement json reference resolver

andy840119:
It's a attempt to resolve part of issue #1107
.
Trying to fix two parts:
- reduce duplicated object in json beatmap. parent lyric in the first note will record as a lyric instance, not reference id.
- due to the first issue, it will cause reference error in parent lyric property on the note class.


andy840119:
Still not fixing due to the bug in `Newtonsoft.Json`.
Something like https://github.com/JamesNK/Newtonsoft.Json/issues/124
.
solution(maybe):
https://stackoverflow.com/questions/25853407/json-net-not-respecting-preservereferenceshandling-on-deserialization
demo: 
https://dotnetfiddle.net/j1Qhu6

andy840119:
Also, should be better not to touch global json config.
So it's ok not fix fix the first part.

andy840119:
This issue can be closed due to relation will be mapping in the beatmap processor.

PR #1157, opened 2022-03-06
Fix submit lyric property value failed in some cases.

andy840119:
image
Fix change ruby, romaji, or note text failed if click to another lyric before submit.


andy840119:
Note: 
it will have the issue that should click two times of arrow key if focus to the cell.
guess the better way might be:
- let the textbox focus-out event be triggered first.
- then, move the caret position.

andy840119:
or should force submit the text first before changing the caret?
maybe should add another event before changing the caret.

PR #1166, opened 2022-03-08
Fix the recording caret algorithm might calculate wrong position if time-tag is not order in the lyric.

andy840119:
Trying to fix the case that time-tag might not always be by order in the list, which will cause the caret algorithm to get the wrong position.


andy840119:
確認一下:
- lyric editor 中的 time-tag 是根據 list 順序還是畫面順序進行繪製?
- 另外，會不會影響到多個 time-tag 在同一個 index 的情況?

或許比較好的方式是在新增 time-tag 時還是讓 list 插入在正確位置?

Issue #1171, opened 2022-03-09
Should be better to auto-focus to the first ruby/romaji tag if change to different lyric.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/157457949-0a22ac63-b3e2-4b98-bb2d-799042162932.png)
It will be better if focus on the input cell on the right side because it will let the user edit the text without moving the mouse.
Use can use the tab and up/down button to change the target lyric and time tag.

andy840119:
Implemented.

andy840119:
Seems there's no need for the config now.

PR #1173, opened 2022-03-10
Should be better to auto-focus to the first ruby/romaji tag if change to different lyric.

andy840119:
Trying to resolve issue #1171.
But will cause cannot change lyric until comment out those lines in the `LabelledObjectFieldTextBox`
```csharp
// should change preview text box if selected ruby/romaji changed.
OnCommit += (sender, _) =>
{
    // because selected lyric might changed if user click another lyric before on commit triggered.
    // so should force to select the lyric back.
    //if (lyricCaretState.BindableCaretPosition.Value.Lyric != lyric)
    //    lyricCaretState.MoveCaretToTargetPosition(lyric);

    ApplyValue(item, sender.Text);
};
```

And those two lines are added for preventing text cannot submit in the #1157


andy840119:
Works perfectly after #1179 merged.

PR #1186, opened 2022-03-13
Fix apply lyric property broken at some place.

andy840119:
Fix after #1185, change lyric properties will be broken after changed to another editing mode.
.
Should re-add the lyric into selecting hit object after the end of the selection.
Those kinds of errors do not figure out before is because there's not able to change the property in the create mode.


andy840119:
Should place the logic into better place, but seems there's no other better solution now.

Issue #1172, opened 2022-03-09
Use keyboard to change the start/end index in the ruby/romaji tag.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/157459282-18b8a82d-dc7b-4a54-834f-d031c97fd513.png)
I think it should be better to let users change the index of the ruby/romaji index directly after fixing the text.

Maybe:
- press `??(function key)` and direction key to change the start index.
- press `??(function key)` and direction key to change the start index.

andy840119:
It might be super hard to find the two empty function keys....

andy840119:
![image](https://user-images.githubusercontent.com/9100368/157465499-c24651f3-4d9d-4b0f-bdda-40efc744dc18.png)
Because text input eats too many keys, so should have an adjustment area on the bottom of the textbox.
And it's focusable.
,
Also, might have the option to:
- only focus on the input area.
- only focus on the change index area.
- focus on the both parts.

andy840119:
Mark as part of #1169

andy840119:
Add change start/end index component in the `LabelledTextTagTextBox`
override `OnFocus` to `OnFocusLost` to update focus state display.

Use `TabbableContainer<T>` to get the tabbed event.
`ITabbableContainer` is used to determine tabbable or not.

andy840119:
Implemented.

PR #1189, opened 2022-03-15
Use keyboard to change the start/end index in the ruby/romaji tag.

andy840119:
Closes issue #1172
.
image
What's done in this PR:
- add some utils for checking the ruby/romaji tag index.
- add more keys for edit the tag start/end index.
- should make sure that the tag is valid in the change handler.
- Implemented text-tag edit area on the button of the edit text area.


andy840119:
i don't like the `xxx(xx):` prefix because it seems ugly and meaningless in this project.

andy840119:
image
Seems works nicely in the romaji mode also.

PR #1218, opened 2022-03-20
Implement switch create time tag caret algorithm in the edit time-tag mode.

andy840119:
This is part of the implementation of #1213 (see: https://github.com/karaoke-dev/karaoke/discussions/1213#discussioncomment-2395884).
Notice that it will use multiple caret algorithms(`TimeTagIndexCaretPositionAlgorithm` and `TimeTagCaretPositionAlgorithm`) in the create time-tag mode, so some of the architecture will be changed.

What's done in this PR:
- create the adjusted action keys(those keys do not implement their feature in this PR).
- add the config to record using which algorithm in the create time-tag mode.
- switch the algorithm in the caret state.
- implement the component to switch that state.
- use algorithm instead of the edit mode in the caret layer for creating the caret and deciding the return class after clicked.


andy840119:
image
Got an issue that cannot delete the time-tag if switch to adjust time-tag mode.
But seems it's still OK to give it a merge.

andy840119:
image
Wait.
The new design will be updated.

Issue #1214, opened 2022-03-20
Fix language dialog not able to select the language right now.


andy840119:
https://streamable.com/6zms35
https://streamable.com/d8aqom

Language selection dialog is still needed to find a suitable place to place that
Because Lazer does intend others to create their own dialog right now.
Maybe:
1. will change the language selection dialog into popover(but it will have the issue that cannot open two popovers at the same time.)
2. wait until Lazer to support customized dialog. 

Or before waiting until Lazer supports customized dialog, it can be:
- make all the dialog(e.g: singer info or language selection) be injectable.
- get the `OsuGame` in the base dialog class.
- if the dialog is destroyed, removed from the modal container.
- it will be not possible to show the modal, but it might be OK right now.

andy840119:
All the dialog that need to be fixed:
- EditSingerDialog : PopupFocusedOverlayContainer
- ManageLayoutsDialog : TitleFocusedOverlayContainer (should be removed.)
- FontSelectionDialog : TitleFocusedOverlayContainer
- LanguageSelectionDialog : TitleFocusedOverlayContainer

andy840119:
Might be separated into two parts:
1. move `LanguageSelector` out of the `LanguageSelectionDialog`
2. change `LanguageSelectionDialog` to popover eventually.

![image](https://user-images.githubusercontent.com/9100368/159720571-83036a6a-5315-43a1-a9a7-fc8f5534fc28.png)
And note that there are lots of places using the `LanguageSelectionDialog`, so it might be hard to do all the things in a single pull request.

andy840119:
Also, after change `LanguageSelectionDialog` to popover, should remove `TitleFocusedOverlayContainer`.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/159934333-1b63c779-ff2e-4ade-86f3-1c937d45b52e.png)
The last question:
- How to add the new language inside this shit.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/159940778-f38425ac-1f5e-4afb-a87c-2a8727c9339d.png)
Guess might make something like that?
Use should add a new language by searching.
And the list of languages will on the top.

andy840119:
Why this shit need that much PRs?

PR #1229, opened 2022-03-24
Replace language selection dialog into popover in some of the pages.

andy840119:
Implement part of #1214

What's done in this PR:
- Fix the behavior in the `LanguageSelector`
- Remove method for placing the dialog inside `osuGame` because it's not needed anymore.
- Implement the popover for the editor page only(Because there are some of the pages that need this popover).
- Change the wording in the delete language popup dialog.
- Remove ask to add first language logic.
- Move some of the components into the individual class.

Those pages are using popover language selector instead of dialog:
- Config page.
- Change the lyric language area in the lyric editor.
- Translate page.


andy840119:
Some of the `LanguageSelectionDialog` are not removed due to having UX issues.
Need to have a discussion.

andy840119:
All the CI service not working 😢

PR #1234, opened 2022-03-26
Remove language selection dialog using.

### LINKED ISSUE #1214: Fix language dialog not able to select the language right now.

andy840119:
Guess it's the final step to close #1214.

What's done in this PR:
- fix the typo.
- rewrite the select language component.
- remove the select language dialog and relative using.


andy840119:
Before:
image

After:
image

andy840119:
Ignore the code factor.

Issue #1169, opened 2022-03-08
The all needed way to adjust the time-tag


andy840119:
should check if should have those features:
- [ ] able to create time-tag in the target of string index.
- [ ] able to remove time-tag in the target of string index.
- [ ] able to move like recording time-tag caret and remove unneeded time-tags
- [ ] able to change the time-tag state (e.g: 5, start -> 4, end)
- [ ] able to shift the time-tag (e.g: 5, start -> 6, start)

andy840119:
![image](https://user-images.githubusercontent.com/9100368/157264192-bcb85657-159a-405b-91bc-323ef9672905.png)
Also, should think about:
- implement all features on the single page or multi-page?

andy840119:
We need to separate into two parts:
- create
  - create the time-tags
- adjust
  - delete the time-tags -> maybe clear `D` or `Delete`
  - shifting the time-tag index or state (maybe press `z + <- or -> ` to change the index, press `x + <- or ->` to change the state)
  - clear the time(maybe? click `R` ?)

andy840119:
Maybe we should create a component to let users able to switch between creating and adjusting mode.

PR #1241, opened 2022-03-27
Implement shifting the time tag.

andy840119:
Guess it's the final step of #1169 and #1213

What's done in this PR:
- Move the get previous/next time-tag index position into utils.
- Add a new method in the change handler to shift the time tag.
- implement receive the shifting time-tag action.


andy840119:
todo: 
- [x] need to test in the adjust time-tag mode.

andy840119:
Although it still has some issues but guesses it's ok to be merged.

andy840119:
fatty

PR #1249, opened 2022-03-31
Move clear selected hitobject in the base class.

andy840119:
For safety purposes.


andy840119:
seems the to-do list CI is not here.

andy840119:
Guess it's time to reduce the test cases

PR #1251, opened 2022-04-01
Fix the caret position is not by order after dragged.

andy840119:
Remove the order utils because the lyric in the list already in order.
Also, it will cause wrong order because reset algorithm will be earlier than the change the order property.


andy840119:
Nice refactor 👍🏼

PR #1256, opened 2022-04-02
Fix back button issue in the import lyric screen

andy840119:
What's fixed in this PR:
- Remove the back button because it will hide the lyric editor.
- Should exit import screen only if in the drag screen(first import lyric step)
- Add abort import feature in the import manager.
- Checking able to roll-back to the target page instead of checking current page if user click the back button.


andy840119:
Also, there's something found but not fixed in this PR(might create another issue for it):
- Should be better to press `Esc` to go-back to previous screen.
- Should think about should show the popover if change the step by fixing invalid part by lyric maker.

Issue #1265, opened 2022-04-07
Interact with localizaton


andy840119:
Not really sure if other rulesets already contain translation...

What's should be done if wants to add the translation:
- Add the `Localisation` namespace and add the testing translate.
- See `ResourceManagerLocalisationStore`, need to check about is there any way to apply the localization for ruleset.
- Check how to generate the translation source file.
- See the project about how to make pull-request if has a new translation https://github.com/ppy/osu-resources.

andy840119:
Can be separated into

Config:
- ConfigSettingsStrings
- StyleSettingsStrings
- ScoringSettingsStrings

Gameplay:
- ModeStrings(Not sure)
- GameplayStrings(Not sure)

Edit:
- LyricEditorStrings(Not sure)
- SingerEditorStrings(Not sure)
- TranslateEditorStrings(Not sure)
- EditorSetupStrings(Not sure)
- ComposerSetupString(Not sure)
- [Note] should consider about some common string, e.g: tooltip or dropdown.

smoogipoo:
I would be interested if you are able to use https://github.com/ppy/osu-localisation-analyser. It's made for this purpose, but hasn't been tested outside of osu! yet.

Add it to your project via both:
https://github.com/ppy/osu/blob/d256bd2cc793d0ea296eea9a75ab09aa5fd92c63/osu.Game/osu.Game.csproj#L33-L36
https://github.com/ppy/osu/blob/d256bd2cc793d0ea296eea9a75ab09aa5fd92c63/.config/dotnet-tools.json#L29-L34

The first tool gives you code refactoring options at the start of non-verbatim strings (strings which don't start with `@`) in tools like Rider, VSCode and Visual Studio. It automatically generates the `XStrings.cs` files for you.
![image](https://user-images.githubusercontent.com/1329837/162358753-c1a4c6ab-2851-41e9-bd6b-83f465972287.png)

The second tool is used to generate the `.resx` files (run `dotnet tool restore` first).
- You can either place these alongside the .cs files via:
  `dotnet localisation to-resx /path/to/project.csproj`
- Or you can redirect them to another directory via:
  `dotnet localisation to-resx /path/to/project.csproj --output /destination/directory/path`
[...]

PR #1266, opened 2022-04-08
Interact with localizaton

andy840119:
It's the base implementation of trying to translate the karaoke ruleset.
The full step of adding localization can see #1265

What's done in this PR:
- Install package for localization hint and auto-generate the class.
- Add dotnet tool.json for able to generate the localization resource file(.resx)
- Create the localization class for testing purposes.
- Create the localization resource file.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "ppy.localisationanalyser.tools": {
+      "version": "2022.320.0",
+      "commands": [
+        "localisation"
+      ]
+    }
+  }
+}
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -10,6 +10,7 @@
 /Directory.Build.props  @andy840119
 /appveyor.yml           @andy840119
 /cake.config            @andy840119
+/crowdin.yml            @andy840119
 
 # Resource related
 /assets/                                      @andy840119
@@ -17,6 +18,7 @@
 /osu.Game.Rulesets.Karaoke.Tests/Resources/   @andy840119
 
 # Editor or git config
+/.config/dotnet-tools.json                    @andy840119
 /.editorconfig                                @andy840119
 /.gitattributes                               @andy840119
 /.gitignore                                   @andy840119
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,0 +1,14 @@
+preserve_hierarchy: true
+files:
+  - source: /osu.Game.Rulesets.Karaoke/Localisation/*.resx
+    translation: /osu.Game.Rulesets.Karaoke/Localisation/%file_name%.%locale%.%file_extension%
+    ignore:
+      # note that this should probably be using %locale% rather than a simple wildcard
+      # but it doesn't seem to respect our mapping below.
+      - /osu.Game.Rulesets.Karaoke/Localisation/%file_name%.*.%file_extension%
+    update_option: update_as_unapproved
+    languages_mapping:
+      locale:
+        ja: ja
+        zh-CN: zh
+        zh-TW: zh-TW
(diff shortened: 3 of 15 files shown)


andy840119:
Let's test on the release build

andy840119:
image
Seems not working in the release build.
Need to check the reason.

[update]
image
Got no idea why `zh.resx` and `ja.resx` is missing while packing.
Checked by the dotpeek.

[another update]
image
Got no idea why the resource is still missing in the local debug build, but Japanese localization can be loaded in the `osuGame` test scene.

andy840119:
Also, not really sure should place the localization file into:
- `<root>/Localisation`  -> will let the code and file in the same namespace.
- `<root>/Resources/Localisation` -> need to setup `.editorconfig` and deal with `Duplicated resource` issue(NETSDK1022)

andy840119:
Make the reference:
https://crowdin.com/project/karaoke-dev
The translation is available here.

andy840119:
Check:
- [x] is all language can be found in the resources dll

image
Damn, got no idea why some of the localization files are missing(or not visible?)

andy840119:
image
But it's OK to catch the fall-back localization fie(?) in the release environment.
So seems it might have an issue getting another kind of localization file?

andy840119:
todo:
- [x] Check all the localization strings if exist. See: https://stackoverflow.com/a/2041057

[update]
Tested in this code, but seems it's only able to read the resource file.
```
 new StreamReader(matchingAssembly.GetManifestResourceStream("osu.Game.Rulesets.Karaoke.Localisation.KaraokeSettingsSubsection.resources")).ReadToEnd()
```

andy840119:
Seems it's the issue related to the `return manager.GetString(key, EffectiveCulture);` in `ResourceManagerLocalisationStore`. It's only able to get the default localization.
Or the localization does not pack well in the release build.
Very not sure.

Not sure related to those kinds of issues:
https://github.com/xamarin/xamarin-android/issues/4664
https://stackoverflow.com/questions/33274715/wrong-culture-string-from-resourcemanager-getstring

smoogipoo:
Hmm.. So it looks like you need the `osu.Game.Rulesets.Karaoke.resources.dll` files placed inside `/bin/{Debug/Release}/netstandard2.1/{locale}/`. Copying them inside the `osu.Desktop/bin/Debug/net6.0/{locale}` dirs seems to work for me.

Can you confirm this is the case?

Bundling the localisation files alongside the ruleset also seems to work, like:
image

This would be a .NET `ResourceManager` localisation, it needs those files and can't search inside the parenting assembly... Not sure how to handle this yet but it's a good find. I would say that bundling the localisation files as above is a reasonable solution for now.

andy840119:
Big thanks for the testing for this issue 🙏🏼  .
I might test this issue in a few days in the osx.

andy840119:
Tested, both way works in the osx.

But have some (maybe) bad news:
- I have made traditional chinese(zh-TW) localization as well, but seems the folder(zh-TW) and resource file doesn't generated.
- Trying to packing them as single dll, but this way seems doesn't work.(Checked that the resource is inside the dll)
image

smoogipoo:
> but seems the folder(zh-TW) and resource file doesn't generated

This is weird. It looks like it needs to be called `zh-Hant` for it to be generated. I've made an issue tracking this here: https://github.com/ppy/osu-localisation-analyser/issues/44.

> Trying to packing them as single dll, but this way seems doesn't work

Yeah, that's why I mentioned `ResourceManager` doesn't search the parenting assembly. I noticed you're using IL repack :smile:

andy840119:
Conclusion:
- Will working if copy the `osu.Game.Rulesets.Karaoke.resources.dll` into the right position.
- `zh-TW` localization resource file not generated issue is fixing.

Guess it's harmless if merged.
Will open another issue if has another new progress(e.g: `zh-TW` issue or found a new way to deal with the localization or interact with CI).

Also, might open another issue for packing the localization resource file as zip for users who wants to use the localization with release build.

smoogipoo:
After https://github.com/ppy/osu/pull/17785, you'll be able to rename it to zh-Hant

PR #1268, opened 2022-04-12
Add some translation

andy840119:
For testing if crowdin CI integration works.


andy840119:
Also, not really sure should be better to make a CI for checking if user generated the localization resource file?

And make the hint like this:
```
run the comment to generate the localization file:
dotnet tool restore
dotnet localisation to-resx ./osu.Game.Rulesets.Karaoke/osu.Game.Rulesets.Karaoke.csproj
```

PR #1276, opened 2022-04-18
Adjust localization.

andy840119:
Redirect zh-TW alias to zh-Hant.
See #1266


andy840119:
Related the change #1277

Issue #1285, opened 2022-04-23
Refactor the gameplay setting overlay.


andy840119:
What's should be adjusted:
- Should give it a solid color, different overlay might have different colour.
- Might change button colour?
- Maybe adjust the namespace follow how setting screen like?
- Remove that ugly `Add()` usage in the `GeneralSettingOverlay`.

![image](https://user-images.githubusercontent.com/9100368/164892848-6e4f5d47-f8a4-44fe-bd19-c302f47854e1.png)

andy840119:
Fixed.

Issue #152, opened 2020-07-28
Test case affect by exist ruleset.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/88677264-a4f60780-d11f-11ea-9218-c5714f497ddb.png)
Seems like debug lazer will read `ruleset` folder's another ruleset's `.dll`
Need to fix while testing ruleset on release lazer.
.
Place an exist ruleset then try to run test case to see which test case is not assign target ruleset.

andy840119:
This issue can be closed because test case change name in #151

PR #1286, opened 2022-04-24
Refactor the gameplay setting overlay.

andy840119:
Closes issue #1285


andy840119:
Note that font alpha issue is fixed by #152

PR #1292, opened 2022-05-02
Update font package to fix the render issue.

andy840119:
Fix render issue by https://github.com/karaoke-dev/osu-framework-font/pull/169
Also, tested if works in the current project because `karaoke sprite text` only support single shader now.

Before:
image

After:
image

Seems around 10 fps of improvement.


andy840119:
No crash error in the lyric editor.
Nice.

PR #1293, opened 2022-05-05
Fix lyric transformer issue

andy840119:
Upgrade the package to fix issue like https://github.com/karaoke-dev/osu-framework-font/pull/177


andy840119:
image
Guess still not fixing that because change the shader not re-calculate the transforms.
Not really sure why not happened in the lyric editor.

andy840119:
But it's still OK to give it a merge.
Seems no feature breaking.

andy840119:
Will make a new release until this issue has been fixed.

PR #1298, opened 2022-05-09
Update package to the latest

andy840119:
Update package to the latest fix the api change.


andy840119:
Seems there are some injection stuff changed.

See:
https://github.com/ppy/osu-framework/pull/5151/files

andy840119:
LGTM!
Thanks for the effort.

Issue #1326, opened 2022-05-15
Fix language detection might be wrong if select all lyrics at the same time.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/168474190-474f5a91-3195-49dd-b441-8711049365a3.png)
All get wrong language if select all the language in the `TestSceneKaraokeEditor`.

![image](https://user-images.githubusercontent.com/9100368/168474242-50ed94b5-1ec8-48ff-b736-fed14f927caa.png)
But the language detection might be right if only select one or two lyrics.

andy840119:
Got the root cause:
Changing the language in the `LyricLanguageChangeHandler` will also trigger `BindableValueChanged()` in the `LanguageInfo`.
Not really sure how to prevent those kinds of issues.

Solution:
- `LanguageInfo` should not bind the property in the lyric.
- Add a flag in the `HitObjectChangeHandler<T>` to check if changing or not.

andy840119:
test.

andy840119:
Implemented.

PR #1327, opened 2022-05-15
Make lyric selection state strategy clear.

andy840119:
Figured out while implementing the #1326

What's done in this PR:
- make sure should check the disable lyrics while adding the lyric to the selected list.
- make a check that not add or remove the selecting list if not selecting.
- should clear all selection after end selecting.


andy840119:
Related to #1326 is because should not add lyric to selected hit object by using this state.

Issue #1312, opened 2022-05-13
Fix the time-tag position


andy840119:
![image](https://user-images.githubusercontent.com/9100368/168328240-3f5408f8-1c97-48ff-8324-720ebccf09c0.png)
Should not get time-tag position by text rect.
Guess should caused by #1293

andy840119:
It need a framework fix also because not able to get the character y offset.

andy840119:
Fixed.

PR #1329, opened 2022-05-16
Upgrade the font package to the latest.

andy840119:
Upgrade the font package to the latest to fix issue #1312 and https://github.com/karaoke-dev/osu-framework-font/issues/193


andy840119:
Tested and make sure that both issue fixed.

PR #1339, opened 2022-05-27
Switch the lrc parser

andy840119:
It's time to switch to `LrcParser`.
It's a re-write package focus on encode or decode the `.lrc` format file.
More test case in this package so should be more stable.
And code quality is much more higher.

Repo:
- https://github.com/karaoke-dev/LrcParser

Nuget:
- https://www.nuget.org/packages/LrcParser


andy840119:
Nice.

Issue #1356, opened 2022-06-02
Implement auto-generator change handler for lyric. 


andy840119:
Inspired from #1335 while implementing this issue.
Not really sure will it be a good idea to make the auto-generated related things into individual change handler.
Maybe named `LyricAutoGenerateChangeHandler`?

What's might be done in this issue:
- [x] implement the change handler and interface.
- [x] Cancel inherit `IAutoGenerateChangeHandler` in other place.
- [x] Add the enum for able to decide generate part(e.g: lyric or ruby-tag).
- [x] Remove the `KaraokeRulesetEditGeneratorConfigManager` injection in the change handler test case.
- [x] should fill the enum in the `AutoGenerateSubsection`.

andy840119:
Also:
- [x] Need to make sure that all the selector and generator should be generate in the real time, for able to get the latest config.

andy840119:
Implemented.

Issue #1335, opened 2022-05-22
Should not able to select if there's no auto-generator for the current language.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/169699296-ccca637c-a5e2-45d4-b067-70f039cef2ce.png)
Should apply those roles into:
- Ruby auto generator.
- Romaji auto generator.
- Time-tag auto generator.

![image](https://user-images.githubusercontent.com/9100368/169699431-9d591049-f3b1-4c20-a3e8-bb6ba0db300e.png)
Also, should show the not support message here.

andy840119:
The final step of this issue might let the generator able to return the invalid message.
Maybe make the `GetInvalidMessage()` interface in the `ILyricPropertyDetector` and `ILyricPropertyGenerator` to replace the exist interface (e.g: `CanGenerate`)

andy840119:
Also:
- [x] Not really sure should add the translate check in the test case? -> not now because it's hard to get the translation key.
- [ ] Should fill the test case in the `LyricAutoGenerateChangeHandlerTest`.

The remain part of this issue might be refactor the test case because contains the same interface.

andy840119:
Guess can close this issue not because add more testing in the `LyricAutoGenerateChangeHandlerTest` is not that really helpful.

PR #1357, opened 2022-06-03
Implement auto-generator change handler for lyric.

andy840119:
Closes issue #1356.


andy840119:
Mark that it's the preparation of the #1335.

andy840119:
Although missing the test case in the `LyricAutoGenerateChangeHandlerTest`, but seems it's OK to make a merge.

PR #1365, opened 2022-06-05
[POC]Add the translation string.

andy840119:
As this PR, it can see that it will create 5 class for the translations, and some of them are duplicated.

Not really sure will it be a good idea or the standard to collect the translations:
- Merge some localization file in a single class, like `GeneratorStrings`.
- Give them a namespace, like `Localisation/Edit/Generator/NoteGeneratorStrings.cs`?
- Common string in the sub-folder? Like `Localisation/Edit/EditCommonStrings.cs`.

### DIFF
--- a/osu.Game.Rulesets.Karaoke/Edit/Generator/Languages/LanguageDetector.cs
+++ b/osu.Game.Rulesets.Karaoke/Edit/Generator/Languages/LanguageDetector.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using osu.Framework.Localisation;
 using osu.Game.Rulesets.Karaoke.Edit.Generator.Types;
 using osu.Game.Rulesets.Karaoke.Objects;
+using osu.Game.Rulesets.Karaoke.Localisation;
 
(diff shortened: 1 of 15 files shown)


andy840119:
Oops, seems placing the localization class into folder cannot be detected by the localization tool.
```
dotnet localisation to-resx ./osu.Game.Rulesets.Karaoke/osu.Game.Rulesets.Karaoke.csproj
```

andy840119:
- [x] test todo list in the description.

CloneWith:
Is it possible to put **all** strings into one class, like `Localisation.UIStrings`?

Then all string translations can use the same method to be fetched, and duplicate strings would be naturally merged into one string.

andy840119:
> Is it possible to put all strings into one class, like Localisation.UIStrings?

No, because will be hard to manage all the translation.

https://github.com/karaoke-dev/karaoke/issues/1265#issuecomment-1146827918

Here's the rule about how to place the translation designed before, i guess will follow this rule after.

PR #1379, opened 2022-06-12
Use current value instead of set value in the component.

andy840119:
What's changed in this PR:
- change event param meaning in the `ObjectFieldSwitchButton`.
- Use current value instead of set value in the component.


andy840119:
現在的 CI 齁
工作量一多就變得慢吞吞的

Issue #1377, opened 2022-06-11
Implement the singer avatar for able to load the image.


andy840119:
See:
1. how this area load image (`BeatmapBackgroundSprite` is load the texture from `IWorkingBeatmap` directly)
![image](https://user-images.githubusercontent.com/9100368/173188681-84e7c8ed-0f1f-48d4-bf35-004d1e85d194.png)
2. how to save the image into beatmap?

andy840119:
and note how to save the file into beatmap:
1. we have GetStream() in the `IWorkingBeatmap`, so technically, it should be able to get the image from beatmap and will be able to convert into the texture.
2. see how `FileChooserLabelledTextBox` to select the file.
3. see how `ResourcesSection` to call the section to change the file
4. seems its use the `BeatmapManager` to add the file to the beatmap.

how to get the image from the working beatmap:
1. see `GetBackground()` in the `BeatmapManagerWorkingBeatmap`. Note that `Background` in the `IWorkingBeatmap` will get the new texture instance every time.
2. Think about how to get the `LargeTextureStore` in the `WorkingBeatmapCache`.
3. Notice that should get the same `LargeTextureStore` instance in the `WorkingBeatmapCache`  in everywhere because it will cache the image.
4. see the `var userResources = new RealmFileStore(realm, storage).Store;` in the `BeatmapManager`. I guess it's the resources to store the beatmap resources.

Guess it will need to create karoake's own service for cache the resource file.
Maybe named `IWorkingBeatmapResources` or `IKaraokeBeatmapResource`, for able to get:
- Singer avatar image.
- Vocal track and BGM track if have(might not implement in the V1).

andy840119:
So seems need to add more to add the image by path to the beatmap in the `SingersChangeHandler`.
And notice that we are missing the test case for this class.

andy840119:
And we need to inject the `IWorkingBeatmap` in the `DrawableSingerAvatar` for able to get the texture from beatmap.

andy840119:
誒不是，我的假日呢？

andy840119:
Implemented.

PR #1380, opened 2022-06-13
Implement the singer avatar for able to load the image.

andy840119:
Closes issue of #1377

What's done in this PR:
- Remove un-need bindable properties because avatar will change the image automatically if avatar changed.
- Remove the `Avatar` to the `AvatarFile` because the property is used to record the upload file name.
- Able to upload the image in the singer change handler.
- Implement singer avatar for able to upload the image by drag or clicking the avatar icon.
- Implement karaoke resource provider for able to get resource from the karaoke beatmap.
- Should be able to re-load the image after update the avatar.


andy840119:
image
Avatar will change after upload the image.

image
Image will be update in another editor also.

image
And the uploaded image is able to be exported.

andy840119:
The remain part is not able to get the `IKaraokeBeatmapResourcesProvider` injection in the `SingerToolTip`.
But i guess it should be OK to fix that shit later.

Issue #1391, opened 2022-06-19
Rename from `Saiten` to `Scoring`.


andy840119:
Because `Saiten(採點)` is from Japanese and will be hard to understand for those who does not understand Japanese.

andy840119:
Implemented.

PR #1392, opened 2022-06-19
Rename from `Saiten` to `Scoring`.

andy840119:
Closes issue #1391.


andy840119:
Seems there's no naming problem with this PR.

PR #1408, opened 2022-06-25
Enable NRT in the utils.

andy840119:
Notice that some of the utils does not handle or accept null value anymore.


andy840119:
Use this tag because laze to create another tag for internal api breaking change.

Issue #1411, opened 2022-06-26
Remove all `AllowNull` attribute eventually.


andy840119:
for now, because not really sure will all BDL property will become like this in the final:
```
[Resolved]
private GameHost host { get; set; } = null!;
```

So add the `AllowNull` attribute for prevent compile error.
And it's easier for searching.
```
[Resolved, AllowNull]
private GameHost host { get; set; }
```

andy840119:
Implemented in the #1988.

PR #1412, opened 2022-06-26
Enable NRT in the skinning.

andy840119:
As title.


andy840119:
Also, mention #1411 because add the `AllowNull` attribute in this PR.

Issue #1417, opened 2022-06-26
Add more test for the convert type extension.


andy840119:
Should add more test case for the following extensions:
- RegexExtensions.GetGroupValue
- TypeUtils.ChangeType

Should test with following types:
- int
- int?
- double
- double?
- float
- float?
- string
- string?

with null or non-null or empty object/string.

PR #1418, opened 2022-06-26
Enable NRT in the extensions.

andy840119:
As title.
Also, should fix the issue #1417 because found some conversion issue in the string.
Should give it a full test eventually.


andy840119:
image
Where's my relaxing time?

Issue #1446, opened 2022-07-18
Should be able to copy the lyric and it's property.


andy840119:
Not really sure will be able to save into the clipboard

andy840119:
Maybe see how other ruleset did?

andy840119:
This is how editor did in the osu ruleset
It will copy the object start time and some of the index into the clipboard.
```
00:21:685 (1,2,4) - 
```

[Update]
`ComposeScreen.cs` does not actually get the state from the clipboard, the selected object is just save as string in the compose screen.

andy840119:
Maybe see how `PerformCopy()` and `PerformPaste()` work in the `ComposeScreen.cs`

andy840119:
Should think about:
1. is copy/paste feature should work in the lyric editor only? (Guess maybe)
2. Should the copied lyric saved in the clipboard or just inside the lyric editor?
3. Guess we will copy the full property in the lyric, and paste the property decide by mode right? 
    - For example, we can copy the lyric from the manage lyric mode, then paste it form the singer mode. It will replace the singer from the copied lyric.

Extra:
1. Will it be a benefit that user can paste the text, then edit it?
2. Will it be a benefit that user can paste .lrc format or text format?

andy840119:
After #1449, guess we should clear the clipboard content if switch the mode?
For prevent paste the wrong json format.

Also, it's there's a need to get the text from clipboard? because it might contains the word that are not expected.

andy840119:
And list all the required apis from the change handler(cut):
- remove all selected time-tag

And list all the required apis from the change handler(paste):
- (optional): replace the text to the lyric(will remove the ruby and romaji)
- paste the list of ruby(might remove the old one)
- paste the list of romaji(might remove the old one)
- paste the list of time-tag(might remove the old one)
- add list of singers.

andy840119:
Guess it's OK enough to give it a close.

PR #1452, opened 2022-07-21
Apply the common interface into the singer change handler.

andy840119:
Also add missing method and test case.


andy840119:
Mark as needed part of #1446

Issue #1468, opened 2022-07-24
Languages should be rename to language in the lyric editor 


andy840119:
Just make a memo, code quality part

andy840119:
Fixed.

PR #1484, opened 2022-07-27
Languages should be rename to language in the lyric editor

andy840119:
Closes issue #1468.


andy840119:
image
Oops, seems api server is not happy now.

Issue #1482, opened 2022-07-27
Implement `IDeepCloneable<T>` for some hit-object for better copy experience.


andy840119:
As title.
Another idea is that maybe it should be possible to serializer to json for copy the property.
For able to make sure that alway contains the new instance.

andy840119:
Maybe let the lyric implement `IDeepCloneable<Lyric>` or `ICloneable<T>`

update:
`ICloneable` does not support the generic type.

andy840119:
Also note that all copyable or splitable object should implement this shit.
E.g. `Note`.

andy840119:
Also, write a test case to make sure that cloned object should not have those issue:
- missing property.
- the same instance(except the property  is mark as reference)

andy840119:
Also, should check that all time-tag in the cloned lyric are the same instance.

andy840119:
And note that can simply the split process in the split note/lyric utils.
And not sure will split lyric will also split the note? Guess should be handle in the change hander 🤔

andy840119:
Implemented.

PR #1490, opened 2022-07-30
Implement `IDeepCloneable<T>` for some hit-object for better copy experience.

andy840119:
It's the final way for the issue #1482.

And note that there's several reason for choosing the json serializer as solution for copy the properties:
1. Follow the way how `PerformCopy()` in the `ComposeScreen.cs` did.
2. Easy way for make sure that all properties has been copied.
3. Easy to make sure that all properties has new instance.


andy840119:
Damn, guess need to write a serializer for this shit.

[update]
There's no need to make the convertor now, just check and add this: https://stackoverflow.com/questions/53405088/output-a-json-property-name-with-a-dash-using-an-anonymous-object

[update again]
The final way is using the json convertor because will have the same issue in the save beatmap process.

Issue #1492, opened 2022-07-30
Invest why got the different value of `CultureInfo` if placing it as key of the dictionary.


andy840119:
found while implementing #1490
`en-US` will become `en_us`
`ja-JP` will become `ja_jp`

Guess the reason is the restriction of json string key format.

Maybe the final way is writing the json convertor for the property like `IDictionary<CultureInfo, string>` in the lyric.cs
Also, test case is required.

andy840119:
Fixed.

PR #1493, opened 2022-07-30
Implement lyric translate json convertor.

andy840119:
Closes issue #1492.
The final solution for now is writing the json serializer with key-value format to fix that issue:

The format should be like this one:
```
[
  { 
    "key":1033, // for better safety, not use the LCID as the key.
    "value":"karaoke"
  },
  {
    "key":1041,
    "value":"カラオケ"
  }
]
```


andy840119:
Hmm..
Waste lots of storage size, but seem OK for now.
Maybe will have better format after 🤔

PR #1507, opened 2022-08-10
Switch the draw node by changing the alpha version.

andy840119:
Test issue https://github.com/karaoke-dev/osu-framework-font/pull/247.
Seems there's no GC issue, happy.


andy840119:
image
And compare to the `SingleShaderBufferedDrawable`, there's no empty FrameBuffer in the `MultiShaderBufferedDrawable`

Issue #1510, opened 2022-08-13
Fix cannot import the lyric due to missing depenedency


andy840119:
Here's the log:
```
2022-08-13 15:47:02 [error]: osu.Framework.Allocation.DependencyNotRegisteredException: The type LyricEditorClipboard has a dependency on ILyricLanguageChangeHandler, but the dependency is not registered.
2022-08-13 15:47:02 [error]: at osu.Framework.Allocation.ResolvedAttribute.<>c__DisplayClass7_0.<getDependency>b__0(IReadOnlyDependencyContainer dc)
2022-08-13 15:47:02 [error]: at osu.Framework.Allocation.ResolvedAttribute.<>c__DisplayClass6_1.<CreateActivator>b__2(Object target, IReadOnlyDependencyContainer dc)
2022-08-13 15:47:02 [error]: at osu.Framework.Allocation.ResolvedAttribute.<>c__DisplayClass6_0.<CreateActivator>b__1(Object target, IReadOnlyDependencyContainer dc)
2022-08-13 15:47:02 [error]: at osu.Framework.Allocation.DependencyActivator.activate(Object obj, IReadOnlyDependencyContainer dependencies)
2022-08-13 15:47:02 [error]: at osu.Framework.Graphics.Drawable.load(IFrameBasedClock clock, IReadOnlyDependencyContainer dependencies)
2022-08-13 15:47:02 [error]: at osu.Framework.Graphics.Drawable.Load(IFrameBasedClock clock, IReadOnlyDependencyContainer dependencies, Boolean isDirectAsyncContext)
2022-08-13 15:47:02 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.loadChild(Drawable child)
2022-08-13 15:47:02 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.load(ShaderManager shaders, Nullable`1 cancellation)
2022-08-13 15:47:02 [error]: --- End of stack trace from previous location ---
2022-08-13 15:47:02 [error]: at osu.Framework.Allocation.BackgroundDependencyLoaderAttribute.<>c__DisplayClass6_0.<CreateActivator>b__3(Object target, IReadOnlyDependencyContainer dc)
2022-08-13 15:47:02 [error]: at osu.Framework.Allocation.DependencyActivator.activate(Object obj, IReadOnlyDependencyContainer dependencies)
[...]

PR #1512, opened 2022-08-14
Fix cannot import the lyric due to missing depenedency

andy840119:
Trying to fix issue #1510.
But seems not working in the release build 🤔 .
Need to make the release build to check that.


andy840119:
Anyway, give it a merge.

Issue #1488, opened 2022-07-29
Able to assign the reference lyric in the right side section.


andy840119:
<img width="346" alt="image" src="https://user-images.githubusercontent.com/9100368/181770216-ec1ea60d-89d5-4175-a9c0-4e3b432d8fd0.png">
Instead of clicking the badge, it might be better just let user able to adjust and see the property in the right side.
Maybe follow the naming like the note section?

Section name and the field in the section:
- Reference lyric
  - Area to select the reference lyric, will show the popover or trigger the selection(should implement the single select mode) after click the select button. Also will have the delete button(or un-link chain icon) at the right-side of the select button.
- Properties (will show this section after contains the reference lyric)
  - Switch area
  - Properties of the sync mode and the reference mode(see the concept in the #1444)

andy840119:
Not really sure should make the sub-edit mode in the reference lyric mode?
Because it might be hard to define a default mode(sync or reference) after user mark the reference lyric.

If the user wants to change the sync mode, they can just switch to the edit mode, then click the reference lyric to "take the control" back.

andy840119:
<img width="300" alt="image" src="https://user-images.githubusercontent.com/9100368/181772972-427dbce9-d9af-425c-b9a0-b8eb2c3c89f9.png">

Another concept(sorry for my bad drawing) is:
- make the each mode as the selection
- all the property are in the selection
- if user click the switch button in the selection, means user want to user that mode.

cons:
- not follow the current UX behavior, means it might be broken in the future.

andy840119:
Another concept(might be the final):
Will be able to show the reference mode or sync mode in the lyric selection popover.
We can simply use the reference lyric detect to detect all the lyric in the popover.

If the lyric is not detectable, means it's time to user the reference mode.

But user still able to change the mode.

andy840119:
Guess it should be OK to close.

PR #1517, opened 2022-08-21
Implement select reference lyric popover.

andy840119:
What's done in this PR:
- Implement the lyric selector.
- Implement the labelled lyric selector.
- Implement the section for able to select the reference lyric.


andy840119:
Mark implement part of #1488

Issue #1519, opened 2022-08-21
Implement generic type convertor.


andy840119:
Implement the `GenericTypeConvertor<T>` for collect some common logic in the base class.

Which class will inherit this class:
- [x] ShaderConvertor
- [x] KaraokeSkinMappingRoleConvertor
- [x] KaraokeSkinGroupConvertor
- [x] KaraokeSkinElementConvertor

andy840119:
Close due to everything done.

PR #1520, opened 2022-08-21
Implement generic type convertor.

andy840119:
Implement part of #1519.

What's done in this PR:
- Implement the base class and expose the abstract method.
- Refactor the logic in the serializer/deserializer.
- Remove the local json serialize. This is not a good using.


andy840119:
And mark that `KaraokeSkinContractResolver` might reference by this:
https://stackoverflow.com/a/18548894

PR #1524, opened 2022-08-21
Adjust the badge.

andy840119:
Rename the badge and add the chart.


andy840119:
https://github.com/erikbern/git-of-theseus
Want to add this but seems there's no online version 😢

Issue #1467, opened 2022-07-24
Implement strategy to sync the value from reference lyric.


andy840119:
## Description
Separate part of #1444.
After `lyric B` reference `Lyric A`, if `Lyric A` change the property, `Lyric B` should change the same property also (in the view).

### There are two strategy for this case
1. Change the property from the reference lyric.]
2. Make all the part(e.g: drawable lyric, editor component) get the referenced lyric's property(getter or the bindable)

### The first strategy
Props: 
- there's no need to touch the component part.

Cons: 
- have lots of sync strategy in the change handler.

Solution:
Write some utils like sync method for sync the value.

### The second one
Props: 
- can roll-back to the previous value after remove the referenced lyric, means it's easy to compare the before/after.
- keep the `origin value` from the lyric.

Cons: 
- might have` lots of bindable event` for get the reference lyric change event.
- might be easy to `mis-write` the value from the referenced lyric to selected lyric, need to prevent that behavior(and might be hard to figure out by human).

Solution:
1. Get the referenced lyric `instance`.
2. Get the `property` from the referenced lyric.
3. Get the `*fake bindable property` from the reference lyric by utils. the property will trigger the change event if

All the solution are listed in the miro also, written in chinese:
https://miro.com/app/board/uXjVOjkco-4=/?share_link_id=696610130048

andy840119:
Another solution:
```csharp
// change to the customized bindable
// will send the trigger event if value from the referenced lyric changed.
[JsonIgnore]
public readonly BindableReferenceList<RubyTag> RubyTagsBindable = new();

// the property for save/load the value to json
public IList<RubyTag> RubyTagsRowValue
{
    get => RubyTagsBindable.OriginValue;
    set
    {
        RubyTagsBindable.Clear();
        RubyTagsBindable.AddRange(value);
    }
}

// the property to get the value
[JsonIgnore]
public IList<RubyTag> RubyTags
{
    // get the value from current lyric or referenced lyric.
    get => RubyTagsBindable;
    set
    {
        // might throw the exception if has reference lyric.
        RubyTagsBindable.Clear();
        RubyTagsBindable.AddRange(value);
    }
}
```
Pros: 
- There's no big change in the change handler and component. Just need to remember to block the write action if contains reference lyric.

Cons:
- Let the pure model become messy.

andy840119:
Or follow how https://github.com/ppy/osu/pull/19225/files did:
```
public struct HitObjectProperty<T>
{
[...]

PR #1538, opened 2022-08-29
Implement sync reference lyric properties.

andy840119:
Continuous working on the issue #1467.

What's done in this PR:
- Implement the sync logic inside the lyric.
- Add some test case for it to make sure that every property will be synced.


andy840119:
Damn, seems CI is not happy now.

Issue #1547, opened 2022-09-02
Add methods to check if the field is editable in the utils or config.


andy840119:
Another alternative way for #1542.

Maybe add `Editable(string fieldName)` in the `IReferenceLyricPropertyConfig`(Better)
Or make the utils for that.

andy840119:
And note that should not use the enum because will not create enum for every property in the lyrics and notes.

andy840119:
Also, maybe add the interface for that?
Because both `IReferenceLyricPropertyConfig` and the `Lyric` will inherit this utils.
Maybe `Note` will use this shit also.

andy840119:
Maybe should make something like `HitObjectWritableUtils` in the editor namespace because those logic is only for the editor only.

Also, should write the test case to get all writable property and test the `HitObjectWritableUtils` for prevent forgot to add the check if create the property.

andy840119:
Maybe use name `LockLyricProperty()` instead because we use:
```csharp
if (LockLyricProperty()) {
  throw new InvalidOperationException()
}
```
and
```csharp
if (LockLyricProperty()) {
  // do something to prevent write.
}
```

more.

andy840119:
Implemented.

PR #1548, opened 2022-09-03
Add methods to check if the field is editable in the utils or config.

andy840119:
Closes issue #1547.
Not really sure is it a good idea to write the business logic into the utils but seems OK for now.
At least it has the test case.


andy840119:
Guess should be OK to give it a merge.
And open another PR to make the refactor(expose "property writable" in the change chandler interface.)

Issue #1554, opened 2022-09-03
Implement utils to get is section editable by modes.


andy840119:
Inspired while thinking of issue #1550
We should block those sections if not editable:
- Badge should not be editable.
- Lyric text section should not be editable.
- Lyric extend area should not be editable.
- Property section part should not be editable.

andy840119:
Also, should return the lock type or Localization message?

andy840119:
Implemented.

PR #1555, opened 2022-09-03
Improve the `HitObjectWritableUtils`.

andy840119:
What's done in this PR:
- Should restrict some property not edit in the `isWriteLyricPropertyLockedByState` (Implement the method).
- Able to get the lock reason in the utils.
- Only expose the method that lyric editor only care.


andy840119:
Mark as preparation of #1554.

Issue #1536, opened 2022-08-28
Should block some user action in the `LyricPropertySection`


andy840119:
Inspired from #1533.
Not really sure should make some good way to disable the control in the base class(`LyricPropertySection`).

andy840119:
Mark as part of #1444

andy840119:
Mark as part of #1563.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/188451670-45a7f7e4-83af-415e-9f1a-9b996559be7e.png)
![image](https://user-images.githubusercontent.com/9100368/188451784-a880c11c-6a7b-460a-b851-6ed4fe3dd011.png)
Maybe should make a block message area to block the edit area.

andy840119:
Also:
- Move the focus event into the base class? -> maybe, also, should get the lock version in the fieid also.
- ~~Should this section able to know the current edit mode, for able to check edit property is valid.~~ -> Please don't

andy840119:
The remain part is not let user able to tab to edit.
Maybe:
1. make another virtual function to update the disable state.
2. move the focus event inside that new function.
3. need to check how to block the tabbed event.

andy840119:
Fixed.

PR #1566, opened 2022-09-06
Should block edit in the lyric property section if property is not editable.

andy840119:
Main part of #1536.

Will lock the section in the lyric editor if property is not editable.
image


andy840119:
This shit is good enough, so just give it a merge.

PR #1568, opened 2022-09-07
Fixup remaining block issue.

andy840119:
What's done in this PR:
- Implement the disable function for override and collect the disable change logic.
- Use better way not to receive the click or key event.


andy840119:
Mark as part of #1536.

Issue #1540, opened 2022-08-29
Should not let user able to modify the property by clicking the lyric editor row.


andy840119:
Part of #1444 
Maybe should add new layer before the rows.
Also, should make the badge seems not editable.
And need to think how and where to add the tooltip.

andy840119:
Mark part of #1563.

andy840119:
This issue can be closed after:
- [x] Should clear the blueprint after lyric writeable state change to disabled.

andy840119:
Implemented.

PR #1573, opened 2022-09-09
Should let the lyric looks disabled if lyric property is not editable.

andy840119:
UI part of #1540.
At least should let the row looks not editable.


andy840119:
Should think more about it, but just give it a merge now.

PR #1584, opened 2022-09-12
Implement trigger disable edit effect method and event.

andy840119:
For prepare issue #1540.


andy840119:
Seems appveyor does not want to work free for me ;_;

Issue #1563, opened 2022-09-05
Block the editable area in the lyric editor if property is not editable.


andy840119:
UI part of #1444.

![image](https://user-images.githubusercontent.com/9100368/189538894-6511e9e5-46de-44a6-8730-c73812a92294.png)
https://karaoke-dev.notion.site/Lyric-editor-mode-and-block-area-c45017e171f043b2b66fc5a5a6d8530d
List all the edit mode, parts and edit property in the parts.

andy840119:
Notice an issue while implementing #1590:
We make the state interface for each rows in the #1582, but it will be hard to trigger the event if trigger the change by keyboard

e.g:
- create or remove the time-tag by keyboard, see: `CreateTimeTagActionReceiver` or `RecordTimeTagActionReceiver`.

andy840119:
And it might be a waste if need to check editable before actually change everywhere.
Not really sure should be able to get invalid change event in the change handler instead?

But still need to consider the pre-check case.
e.g: should block the drag feature instead of return value not changeable even after actually drag.
So should not remove the `IsSelectionsLocked()` in the `ILyricPropertyChangeHandler`.

Solution:
1. add the fallback action in every change handler. e.g: `void SetLanguage(CultureInfo? language);` -> `void SetLanguage(CultureInfo? language, Action lockAction = null);`
2. Should have lock action event in every change handlers inherit `HitObjectPropertyChangeHandler`.

PR #1590, opened 2022-09-14
Implement disallow edit effect in the edit lyric row.

andy840119:
Maybe the part of issue #1563.

What's done in this PR:
- should not be able to type or remove the text if change handler does not allow.
- add the basic disallow edit effect.


osu.Game.Rulesets.Karaoke/Edit/Lyrics/Rows/Edit/Carets/DrawableLyricInputCaret.cs
```diff
@@ -60,6 +62,12 @@ public DrawableLyricInputCaret(bool preview)
                     ReleaseFocusOnCommit = false,
                     NewCommitText = text =>
                     {
+                        if (lyricTextChangeHandler.IsSelectionsLocked())
+                        {
+                            editLyricRowState.TriggerDisallowEditEffect();
+                            return;
+                        }
+
```
andy840119, line 70: Maybe need the discussion in the https://github.com/karaoke-dev/karaoke/issues/1563#issuecomment-1246788132 but  seems OK to let it merge now.

Issue #1595, opened 2022-09-17
[Proposal] Create the time-tag expert mode.


andy840119:
The most beneficial for the new layout is edit time-tag(or can called edit timing) mode.
But it might not safe if just touch the exist layout.

Prepare steps:
- Rename the extend area (Implemented in the #1596)
- Rename the lyric row are and move the rearrangeable and selection things into the same namespace.

So:
1. Create Edit time-tag (Expert) mode after exist edit time-tag mode.
![image](https://user-images.githubusercontent.com/9100368/190863118-921da98e-f3ed-4622-8ac3-aa4b9709a76e.png)
2. Refactor the lyric editor to let it accept the new layout.
3. Create all necessary components for the edit time-tag (Expert) mode. It should just named `Expert` or `TimeTagsV2`
![image](https://user-images.githubusercontent.com/9100368/190863249-2fba4d68-76d1-4bf5-bf33-24b1a0667ed4.png)
4. Let's test this shit.
5. Delete the old logic.
6. rename the `TimeTagsV2` folder into `TimeTags` to replace the old mode.

andy840119:
Guess this issue can be closed because new layout has been implemented.

andy840119:
Wait.
This issue does not being implemented.
This issue is focus on making another flow to set the time-tag time.

PR #1605, opened 2022-09-24
Prepare for accept the detail view mode in the lyric ist.

andy840119:
Prepare of issue #1595.
Adjust remaining override things.


andy840119:
All test case passed in the local.

Issue #1617, opened 2022-10-01
Remove the extend area in the lyric editor row.


andy840119:
After Migrate to the #1594 (composer is implemented in the #1616)
There's no case that extend area will appear in the lyric row.
So it's time to remove this shit.

Also, mention #1609 because it will be part of the composer i guess?
Might be named bottom edit area?

andy840119:
Implemented.

PR #1623, opened 2022-10-02
Remove the extend area in the lyric editor row.

andy840119:
Closes issue #1617.

Move the extend area into composer and remove some old logic.


andy840119:
Local test case passed.

Issue #1609, opened 2022-09-26
Implement the compose area in the lyric editor.


andy840119:
Part of issue #1594

![image](https://user-images.githubusercontent.com/9100368/192332792-f06c3370-6a31-4f51-8038-056c64f35816.png)
This is the whole lyric editor.

![image](https://user-images.githubusercontent.com/9100368/192332859-947a7386-89bc-47b3-8757-da311342f177.png)
Here's the compose area.
Open an issue for it because it has many component.

And note that the issue does not contains the whole features in the composer.
Only need to implement the layout part

Here's the todo list:
- [ ] Implement the toolbar and some of the buttons.
- [ ] Implement the lyric property workspace.
- [ ] Implement the invalid message workspace.
- [ ] User should be able to show/hide the workspace if the width is not enough.
- [ ] [Optional] Not really sure should write the state for the composer, or making the config.

andy840119:
Also:
- [x] Move the extend part into individual namespace (#1617).

andy840119:
![image](https://user-images.githubusercontent.com/9100368/192549591-116ff278-3ea9-4da6-8e8f-ccef8e4431e5.png)
Maybe called the invalid lyric part as [workspace](https://www.pro-face.com/otasuke/files/manual/gpproex/new/refer/mergedProjects/starttoquit/starttoquit_mm_editorwindownames.htm) or [panel](https://www.infragistics.com/community/blogs/b/nanil/posts/blend-for-visual-studio-workspace-xaml)?

andy840119:
Guess this topic is tooooo old and can be closed.

PR #1624, opened 2022-10-02
Rename some components in lyric editor composer.

andy840119:
For prepare issue #1609
Should remove some of the "editor" wording in the components.


andy840119:
All test case passed in the local.

PR #1627, opened 2022-10-04
Move single lyric editor into shared namespace.

andy840119:
Prepare text in the #1609
namespace is followed `osu.Game.Rulesets.Catch.Edit.Blueprints.Components.EditablePath`


andy840119:
local test passed.

Issue #1649, opened 2022-10-12
Fix got the error if switch from typing mode into split text mode in the lyric editor.


andy840119:
Guess caused by #1643
Happened because caret and bindable caret does not the same type.

Here's the log:
```
System.InvalidCastException: Specified cast is not valid.
   at osu.Game.Rulesets.Karaoke.Edit.Lyrics.Components.Lyrics.Carets.DrawableCaret`1.<>c__DisplayClass2_0.<load>b__0(ValueChangedEvent`1 e) in D:\Github\karaoke-dev\karaoke\osu.Game.Rulesets.Karaoke\Edit\Lyrics\Components\Lyrics\Carets\DrawableCaret.cs:line 43
   at osu.Framework.Bindables.Bindable`1.TriggerValueChange(T previousValue, Bindable`1 source, Boolean propagateToBindings, Boolean bypassChecks)
   at osu.Framework.Bindables.Bindable`1.SetValue(T previousValue, T value, Boolean bypassChecks, Bindable`1 source)
   at osu.Framework.Bindables.Bindable`1.TriggerValueChange(T previousValue, Bindable`1 source, Boolean propagateToBindings, Boolean bypassChecks)
   at osu.Framework.Bindables.Bindable`1.SetValue(T previousValue, T value, Boolean bypassChecks, Bindable`1 source)
   at osu.Framework.Bindables.Bindable`1.set_Value(T value)
   at osu.Game.Rulesets.Karaoke.Edit.Lyrics.States.LyricCaretState.refreshAlgorithmAndCaretPosition() in D:\Github\karaoke-dev\karaoke\osu.Game.Rulesets.Karaoke\Edit\Lyrics\States\LyricCaretState.cs:line 103
   at osu.Framework.Threading.ScheduledDelegate.InvokeTask()
   at osu.Framework.Threading.ScheduledDelegate.RunTaskInternal()
   at osu.Framework.Threading.Scheduler.Update()
   at osu.Framework.Graphics.Drawable.UpdateSubTree()
   at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c)
   at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
   at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c)
   at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
[...]

PR #1650, opened 2022-10-13
Refactor the drawable caret and drawable part layer.

andy840119:
What's done in this PR:
- Make the code more understandable.
- Refactor the logic in the caret layer.
- Reduce some duplicated bindable. We can just pass the caret position to the caret instead of listen the event inside the drawable caret.
- Fix the issue #1649.


andy840119:
damn test case broken 🤔

Issue #1651, opened 2022-10-13
To restrict the caret that cannot change the lyric by move left or right.


andy840119:
Not really sure should restrict(or implement the behavior) it in the:
1. caret state.
2. implement the algorithm.

Also should notice that:
1. Might implement the feature to move to the first/last time-tag or text position in the target of lyric.
2. Need to think about the behavior if need to do the `1.` with or without strict the the same lyric.

andy840119:
Inspired while implementing the #1610.

andy840119:
Also, we might have the lock config, not really sure should it be a part of algorithm or not?

andy840119:
And maybe should add those two actions into the `MovingCaretAction`:
- FirstOfTheLyric
- LastOfTheLyric.

or:
- FirstInLyric
- LastInLyric

(update)final:
- MoveToFirstIndex
- MoveToLastIndex

see the new naming in the #1654

andy840119:
https://karaoke-dev.notion.site/Caret-023d6d71c9504bab9c51f32d99558814
Document for me to clarify the question and implement step.
Written in chinese.

andy840119:
Guess this issue can be closed.

PR #1653, opened 2022-10-13
Change caret position format from class to structure

andy840119:
Why need this change:
- For better comparation to check if the same caret or not.
- Technically, caret should be the same if all property(lyric and some index) are same. There's no case that we need to separate them.

Also should notice that in #496 we use struct in the `CaretPosition`.
Not really sure way it changed to the class, maybe just forgot to use the struct at that time.

Also, all test case passed and seems no issue while testing it with hand.


andy840119:
Mark as preparation of #1651.

PR #1656, opened 2022-10-14
Refactor caret algorithm (make some property protected and add the caret validator).

andy840119:
What's done in this PR:
- Remove the `public` prefix in the interface.
- Make some method protected. only need to let interface public.
- Implement the `validate()` to check all input and output caret should be valid.


andy840119:
The only known error is that will throw the error if delete the whole text.
need to find a good way to fix that.

PR #1658, opened 2022-10-15
Refactor caret algorithm (adjust base interface)

andy840119:
Should use the `IIndexCaretPosition` in the `IIndexCaretPositionAlgorithm`.


andy840119:
Mark as part of #1651.

PR #1669, opened 2022-10-18
Let change handler able to handle the dragging case.

andy840119:
Because dragging to change the property might call the change handler method every frame, which might cause `UpdateState()` called frequently in the `EditorBeatmap`.

So maybe:
- Use the `PerformOnSelection()` in the `EditorBeatmap` for most cases.
- Just change the `SelectedHitObjects` in the `EditorBeatmap` if the state is between `BeginChange()` and `EndChange()` in the `EditorChangeHandler`.


andy840119:
Need to think should be suitable or not, but technically should be OK I guess(?)

Or should be it be better make two function in the `HitObjectChangeHandler<T>`?
- protected void PerformOnSelectionInEditor<T>(Action<T> action) -> which will call `beatmapProcessor` and `processHitObject` in the editor beatmap.
- protected void PerformOnSelectionByChangeHandler<T>(Action<T> action) -> for the simple change.

andy840119:
Or maybe have the abstract attribute to mark should handle change by editor beatmap or editor change handler?

andy840119:
Just let it merge now.
This issue can decide later.

PR #1673, opened 2022-10-22
Switch to another packing tool

andy840119:
Switch the upload file from the "actions/upload-release-asset@master" into "softprops/action-gh-release@v1" because the github action is not maintained anymore.

See the latest change in the https://github.com/LumpBloom7/sentakki/blob/master/.github/workflows/release.yml


andy840119:
Mark that need to add the "draft" for able to add the text at the top.

andy840119:
And seems able to pass the CI this time.

andy840119:
The old release script is added by #132

Issue #1672, opened 2022-10-22
Fix the `dependency is not registered` issue even if variable is nullable


andy840119:
https://github.com/karaoke-dev/karaoke/blob/15e42b1a41b729e211543ce8e805c369ffe93ace/osu.Game.Rulesets.Karaoke/Edit/ChangeHandlers/HitObjectChangeHandler.cs#L27
This dependency is mark as nullable, means it accept the null case.
But it will got the error only in the release build.

Not really sure it's caused by the packing tool.

here's the log:
```
2022-10-22 03:28:39 [error]: osu.Framework.Allocation.DependencyNotRegisteredException: The type HitObjectChangeHandler<Lyric> has a dependency on IEditorChangeHandler, but the dependency is not registered.
2022-10-22 03:28:39 [error]: at osu.Framework.Allocation.ResolvedAttribute.<>c__DisplayClass7_0.<getDependency>b__0(IReadOnlyDependencyContainer dc)
2022-10-22 03:28:39 [error]: at osu.Framework.Allocation.ResolvedAttribute.<>c__DisplayClass6_1.<CreateActivator>b__2(Object target, IReadOnlyDependencyContainer dc)
2022-10-22 03:28:39 [error]: at osu.Framework.Allocation.ResolvedAttribute.<>c__DisplayClass6_0.<CreateActivator>b__1(Object target, IReadOnlyDependencyContainer dc)
2022-10-22 03:28:39 [error]: at osu.Framework.Allocation.DependencyActivator.activate(Object obj, IReadOnlyDependencyContainer dependencies)
2022-10-22 03:28:39 [error]: at osu.Framework.Allocation.DependencyActivator.activate(Object obj, IReadOnlyDependencyContainer dependencies)
2022-10-22 03:28:39 [error]: at osu.Framework.Allocation.DependencyActivator.activate(Object obj, IReadOnlyDependencyContainer dependencies)
2022-10-22 03:28:39 [error]: at osu.Framework.Allocation.DependencyActivator.activate(Object obj, IReadOnlyDependencyContainer dependencies)
2022-10-22 03:28:39 [error]: at osu.Framework.Allocation.DependencyActivator.activate(Object obj, IReadOnlyDependencyContainer dependencies)
2022-10-22 03:28:39 [error]: at osu.Framework.Graphics.Drawable.load(IFrameBasedClock clock, IReadOnlyDependencyContainer dependencies)
[...]

PR #1674, opened 2022-10-22
Create a new folder and copy all reuleset needed dlls in release build.

andy840119:
Because packing tools might have some issues that hard to control(see: #1672)
So we need an alternative way to let user able to download all dlls and placing them into the folder.

See the config in the https://stackoverflow.com/a/44378406/4105113.


andy840119:
Should be OK to let it merge.

PR #1675, opened 2022-10-22
Add more dlls into list

andy840119:
Add some dlls to fix:
- Auto-generate ruby/romaji not working in the editor.
- Localization not showing up.

If need to make the file with folder, then should use `DestinationFiles` to assign the path for each files instead of `DestinationFolder`.

See the document:
https://github.com/MicrosoftDocs/visualstudio-docs/blob/main/docs/msbuild/copy-task.md


andy840119:
Not adding list into the `ILRepack.targets` because it will have the conflict while packing.

PR #1678, opened 2022-10-23
Fix the wrong position calculation in scroll container.

andy840119:
Continuous of #1677.

What's done in this PR:
- Adjust the method location in the `AdjustTimeTagScrollContainer`
- Remove the `Track` instance in the `TimeTagScrollContainer` by following how `Timeline` did in lazer.
- Fix get the wrong snap time in the `AdjustTimeTagScrollContainer`.
- Fix time-tag not blueprint not update well in the `BindableBlueprintContainer<T>`
- Fix release build broken. see: https://github.com/karaoke-dev/karaoke/actions/runs/3305977988

### DIFF
--- a/osu.Game.Rulesets.Karaoke/Edit/Lyrics/BindableBlueprintContainer.cs
+++ b/osu.Game.Rulesets.Karaoke/Edit/Lyrics/BindableBlueprintContainer.cs
@@ -3,6 +3,7 @@
 
 #nullable disable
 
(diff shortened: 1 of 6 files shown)


andy840119:
IDK why test case broken here.
Just merge.

andy840119:
Damn.
The test case seems broken only if run all the tests at the same time.

Issue #1687, opened 2022-10-24
Think the strategy how to get the text-tag or time-tag in the `Issue` class


andy840119:
Needed part of #1685 and #1686

Because we don't want to create the customized `Issue` class for saving the time-tag or the text-tag, so maybe:
- Still remain the class. The benefit is we can group the issue by the issue type.
- Save the time-tag into the last argument. the last argument will not be the part of message string.
- Add the params in the `IssueTemplate`

andy840119:
I guess should be better to add the issue class back.
And the params is focused on letting outside knows the invalid value.

andy840119:
Implemetned.

Issue #1679, opened 2022-10-23
Refactor the checks in the editor.


andy840119:
Here's the link:
https://karaoke-dev.notion.site/Checks-f24020e7ebbc49768a87b40f14ec8f7b

What's might be done:
- Refactor the test case
- Rename the exist checks
- Remove the derived issue class (e.g. `LyricTimeIssue`)
- Add more checks.

andy840119:
Hmm...
Not really sure should write a new table/tooltip to show the invalid message.
Also, might be able to get the icon for each input template.

andy840119:
Also, not really sure should be better to use the class to wrap the `IssueTemplate` with issue type
e.g:
- InternalErrorIssueTemplate

See:
https://karaoke-dev.notion.site/Checks-f24020e7ebbc49768a87b40f14ec8f7b

More about the `InternalErrorIssueTemplate `:
- `IssueTemplate` with `IssueType.Error`
- Add the text like `error code` and `please contact to developer`
- Issue template that able to auto-generate the status code.

andy840119:
Also, it's time to remove the `IHasConfig<T>` in the `osu.Game.Rulesets.Karaoke.Edit.Checks.Configs` until lazer support has config for checks.

andy840119:
Need to check at lyric:
- ID (should not be duplicated)
- Order (Should not be duplicated)

[Done] Need to check at note:
- Text(should not be empty)
- RubyText (should be empty or having the text)
- ReferenceTimeTagIndex (should be able to get the referenced time-tag and should have the time.)

andy840119:
[Done] And note that besides `CheckHitObjectProperty`.
Might be better to create something like `CheckHitObjectReferenceProperty` for better interface naming and using?

andy840119:
[Done] And need to add more checks in the test case:
- PossibleTemplates in checks should not be duplicated.
- All errors should be in the `PossibleTemplates`.

andy840119:
Need to think about:
- How to deal with interface check. e.g: want to check ID or order should not be duplicated in both `Lyric` and `Singer`

andy840119:
Guess should be OK enough to give it a close.

PR #1688, opened 2022-10-25
Add the customized ruby/romaji issue class back.

andy840119:
What's done in this PR:
- Closes issue #1687.
- Roll-back some part removed in the #1684


andy840119:
Mark as relative to the #1679

PR #1690, opened 2022-10-26
Refactor the note check.

andy840119:
Part of #1679.

What's done in this PR:
- Implement the base method to make the check able to access the whole hit-objects in the beatmap.
- Refactor the check for reference lyric in the note.


andy840119:
Finally all test case passed again.

Issue #1696, opened 2022-10-30
Make the flow better between invalid table, caret and blueprint selection.


andy840119:
Mark as part of #1685.

The whole things need to be consider:
- How to trigger the caret or selection if click the invalid table's row.
- Sync the selections/caret between caret and blueprint selection.

Maybe:
1. Add the ruby/romaji caret algorithm back. Not really sure need to implement the move caret event because `tab` will handle that (algorithm is removed in the #1301).
2. Will clear and re-select the selection if caret position changed. Also, should mark the acceptable caret type in the interface or base class
3. Should navigate to the caret if only select one selection. Might implement in the base blueprint class.
4. Prevent recursive call.

andy840119:
Need to think about:
- Need to modify the `TriggerSelect()` in the LabelledObjectFieldTextBox<T>, but caret algorithm does not include the note.
- SelectionItems in the `LabelledObjectFieldTextBox<T>` should be read only.
- How to add the logic in the ModeStateWithBlueprintContainer<T> (mark as related to the #1581)?

andy840119:
Also:
- [x] Need to add the assert check to make sure that caret create outside of the caret algorithm should have the right generated type. -> see: #1706

andy840119:
Also need to think about:
- [x] The new ruby/romaji caret algorithm should replace the old one. -> Implemented in the #1701
  - [x] Should handle the basic switch lyric event.
  - [x] Should focus to the first ruby/romaji if current lyric has ruby/romaji
  - [x] Ruby/Romaji text tag in caret position should be nullable
- [x] Should do the same things to the note caret state. -> implemented in the #1703
  - [ ] Need to think about how to get the notes in the note caret algorithm? 
  - [ ] Should reload the caret algorithm if note changed?

andy840119:
Damn.
Maybe it's not a good idea to sync the caret position and selection item.
Because:
- some of the edit mode we need to focus to the first item if move to different lyric, but sometimes not.
- we will not focus to the first item if change the lyric by click.

andy840119:
Here's the benefit to do or not to do that:
https://karaoke-dev.notion.site/Caret-position-blueprint-6371672d86b643f494aee57c865c694a
All written in chinese.

andy840119:
After discussion, there's no much benefit to sync the selection item between lyric and the selection list.
This is only for making issue table happy.
[...]

PR #1697, opened 2022-10-30
Should make sure the generate type is correct in the base class.

andy840119:
For prevent let the code broken after implement #1696 without notice.


andy840119:
Damn, those failed test cases does not happened every time.

PR #1700, opened 2022-10-31
Implement ruby romaji caret position algorithm.

andy840119:
First step of #1696.
We need to create the ruby/romaji caret position for let all invalid ruby/romaji click can focus to the correct ruby/romaji.


andy840119:
Damn, forgot to think about how to apply to the caret state 😢

PR #1708, opened 2022-11-06
Make edit state able to select target item.

andy840119:
Before decide should implement the #1696 or not
It should be better to collect the logic for select the single blueprint item.

Where will use this method:
- labelled row.
- issue table.


andy840119:
Damn why this shit CI error appear again and again 🤔

Issue #1722, opened 2022-11-12
Implement the issue amount in the verify mode button.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/201475416-107825bd-9fcc-4e27-b3ca-8b1f9a3e3d7d.png)
The base edit mode switch button strategy is added in the #737
It have bindable<int> in the `EditModeSelectionItem<T>`, which is the model class for store the text and pass it into the button:
```
protected class EditModeSelectionItem
{
    /// <summary>
    /// The text which this <see cref="EditModeButton"/> displays.
    /// </summary>
    public readonly Bindable<LocalisableString> Text = new(string.Empty);

    /// <summary>
    /// The description which this <see cref="EditModeButton"/> displays.
    /// </summary>
    public readonly Bindable<DescriptionFormat> Description = new();

    /// <summary>
    /// The alert number which this <see cref="EditModeButton"/> displays.
    /// </summary>
    public readonly Bindable<int> Alert = new();

    public EditModeSelectionItem(LocalisableString text, LocalisableString description)
        : this(text, new DescriptionFormat { Text = description })
    {
    }

    public EditModeSelectionItem(LocalisableString text, DescriptionFormat descriptionFormat)
    {
        Text.Value = text;
        Description.Value = descriptionFormat;
    }
}
```

But it need to write extra logic for able to get the invalid issue amount from the `LyricEditorVerifier`.
Guess should be the time to decide should still use this interface, or design the better one 🤔

andy840119:
What should we done:
1. Remove the bindable in the `EditModeSelectionItem` because technically we will not change the text after assigned.
2. Make the create button method become overridable.
3. Create the button that have number in the top right.

andy840119:
And because it's fucking weird if separate the interface between create the button instance and the button text.
So the interface might be:
- Create action by mode.
- Get the fucking deccription.

Also, should need to think about:
- Should remove the `GetColour()` method? because we already allow to create the button instance 🤔

andy840119:
I guess the final overridable interface might be:
- GetSelectionInstance(); -> have virtual method. Also, the verify button should be abstract and edit mode should be overridable instead of just assign.
- GetSelectionText();
- GetSelectionColour();
- GetSelectionDescription();

andy840119:
Implemented.

PR #1749, opened 2022-11-21
Add missing issue table in texting and reference edit mode.

andy840119:
image
image
Implement the verify section and the issue table for the texting mode and the reference lyric mode.


andy840119:
Mark as part of #1722 because we add the editor check in the #1717 but some of the issue table is missing.

Issue #1754, opened 2022-11-23
Updating project to use File Scoped Namespaces


andy840119:
https://www.meziantou.net/updating-your-project-to-use-file-scoped-namespaces.htm

maybe just run
```
dotnet tool update --global dotnet-format
dotnet format MySolution.sln --severity info --diagnostics=IDE0161
```

Maybe we will upgrade those project to the latest first:
- [x] https://github.com/karaoke-dev/LrcParser -> This package is file scoped namespace already.
- [x] https://github.com/karaoke-dev/osu-framework-font -> Implemented in the https://github.com/karaoke-dev/osu-framework-font/pull/292
- [x] https://github.com/karaoke-dev/osu-framework-microphone -> https://github.com/karaoke-dev/osu-framework-microphone/pull/273
- [x] https://github.com/karaoke-dev/karaoke (test case first) -> Implemented in the #1877.
- [x] main project. -> Implemented in the #1967.

andy840119:
Finally can close this issue.

PR #1756, opened 2022-11-26
Run dotnet format

andy840119:
Run dotnet format and some re-format.


andy840119:
Mark as preparation of #1754.

Issue #1769, opened 2022-11-30
Refactor singe matadata.


andy840119:
Wnat's should be done in this issue:
- [x] Implement the singer info class. -> Implemented in the #1774
  - [x] should be able to get list of main singer
  - [x] [Optional] should be able to get list of sub-singer by main singer.
  - [x] [Optional] should be able to get IDictionary<Singer, SubSinger[]> by (maybe lyric or singer ids, not really sure)
  - [x] class should have `IList<ISinger>`? Not really sure should save by sorting or not?
- [x] Apply singer info into the karaoke beatmap. -> Implemented in the #1775
- [x] Guess should write the json convertor for the generic type.  -> Implemented in the #1771
- [ ] Not really sure should implement the `ICompare<T>` to those singer classes?
- [x] Guess should add `support-sub-singer` flag and should be default as `false`. -> Added in the #1776.
- [ ] Should be able to get the singer instance directly in the singer class.
- [ ] `IList<int> SingersIds` should not have the bindable and should be marked as json ignore.

And notice that did not make unnecessary things for `sub-singer` because not really sure will implement this or not.
Might not be that useful i guess.

andy840119:
Damn, will have those issues:
- We have utils to check if singer instance is in the singer, not really sure should check by singer index or singer instance in the lyric.
- Should separate or combine the lyric should copy the singer instance also?

andy840119:
Not very sure about last part because it should be possible to get the singer instance by working beatmap 🤔

andy840119:
Maybe change to the array and use the validator to check if singer instance is being synced or not

andy840119:
Maybe should make the validator more global?
Reference lyric and singer listance will check by the flag

PR #1773, opened 2022-12-03
Refactor change handler.

andy840119:
For prepare implementing #1769.


andy840119:
Mark as preparation of #1775.

Issue #1767, opened 2022-11-29
Move the screen related things into screen namespace.


andy840119:
Seems `edit` namespace not suitable to place more screen?
Not very sure, just move more things into the screen namespace.

Maybe the structure like:
- Edit 
  - Blueprint
  - ChangeHandler
  - Components(for main editor page only)
- Screen
  - Edit 
    - Beatmaps(include some sub-screen)
      - Lyric editor
      - Page editor
      - Singer editor
      - Translations 
    - Classic layout editor.
    - Scrolling layout editor.
    - Skin
  - Ruleset config

Another rules:
- Should place the logic part (e.g. change handler or auto-generator) into `Edit` namespace.
- Should place the component into `Edit` namespace if use in the main edit screen.

andy840119:
Implemented.

PR #1777, opened 2022-12-04
Move the screen related things into screen namespace.

andy840119:
Closes issue #1767.
For prevent `Edit` namespace become messy.
Should move all new screen related things into the screen namespace.


andy840119:
Damn WTF

Issue #1766, opened 2022-11-27
Implement page (or maybe called breath) edit mode in the lyric editor.


andy840119:
Inspired from #1764

https://youtu.be/alR4Ac9evFg?t=93
Should be useful to create "page mode" in the lyric editor.
Same page means user can sing the lyric with one breath.
Usually between 1~4 lyrics.

Benefit:
- Might let user not confused if they can see the page in the layout editor. It might be useful if user can decide the lyric position by breath.
    - classic layout 可能會根據這個規則進行格式設計
- Will be used in the scrolling layout mode.
- 未來的 layout 可以根據 page 進行參考

andy840119:
需要思考:
- 如果有遇到多個 singer 輪流唱歌詞，可能會遇到 page 跳號的問題，或許應該要有 checks 來檢查這件事情

andy840119:
Maybe can use page number * magic number(e.g. 21) % 360 to get the hue number

andy840119:
Wait
Page can be designed as control point
Should be better to do that?

andy840119:
If add the check:
- [x] Should check by the page interval. -> Implemented in the #1780.
- [x] Should check the same page amount? -> Implemented in the #1780.

andy840119:
Also, should move those classes into more generic location:
- [x] Move the markdown component into editor-level -> Implemented in the #1766.
- [x] `EditModeSection<T>` and `LyricEditorSettings ` , should make another base class for all editor. -> Implemented in the #1785.
- [x] Move the issue table into more general namespace. -> Implemented in the #1787
- [x] Should move the `IssueSection` into more generic location. -> Implemented in the ##1787
- [ ] Screens/Edit/Beatmaps/Lyrics/Settings/Components
  - Should move the block section wrapper into more generic location, those classes should be able to serve the whole editor level.

andy840119:
Should prepare:
- [x] service for able to verify the section. -> implemented in the #1793

Should implement those edit sections:
- [x] show all pages. -> Implemented in the #1797
- [x] Issue table -> implemented in the #1793
- [x] Switch edit mode section -> implemented in the #1793

andy840119:
- [ ] Also, should write the generator for the page editor.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/208276262-818e10f8-2758-4c08-b501-d5175670666b.png)
Need to make some improvements:
[...]

PR #1791, opened 2022-12-11
Implement edit mode state for page edit mode.

andy840119:
Implement part of issue #1766.

What's done in this PR:
- Move the edit mode state interface for more generic location.
- Implement the state provider for the page editor.
- Apply the edit mode switch section to the page editor.


andy840119:
Need to think about should not merge this because:
- [ ] It will feels like interface for UI, not UI use this interface 🤔 
- [ ] This interface is just for the lyric editor only before.
- [ ] It should be ugly if component can get the interface directly.
- [ ] There's no benefit (e.g. performance) to make the interface generic.

Where will use this edit mode component:
- lyric editor.
- page editor.
- classic layout editor.
  - Maybe include other layout editor.

andy840119:
Guess should be better to close this issue for prevent over-design.

PR #1813, opened 2022-12-27
Upgrade package to the latest.

andy840119:
Update the package to the latest.
Will merge until lazer publish the new release.


andy840119:
Guess might not have new release until next lazer release.
So give it a merge.

Issue #1815, opened 2022-12-28
Fix the release build not working.


andy840119:
<img width="649" alt="image" src="https://user-images.githubusercontent.com/9100368/209748760-d697fd2c-be4c-4c64-8093-49de0bfbc58c.png">
Need to change the target folder into `.net6`

Here's the log:
https://github.com/karaoke-dev/karaoke/actions/runs/3791268767/jobs/6446573077

andy840119:
Fixed.

PR #1816, opened 2022-12-28
Fix the release build broken.

andy840119:
Close issue #1815.

Mark as remaining part of issue #1813
Change the target folder into `.net6`


andy840119:
Guess there's no need to wait for the CI.

Issue #1755, opened 2022-11-24
Implement stage class as metadata in the karaoke beatmap


andy840119:
What's will planning:

1. Layout will be the part of metadata in the karaoke beatmap, and will have following type of layouts:
    - Classic -> Current version .
    - Scrolling -> Might appear in the V2.
    - Animation -> Might appear in the V2 or later

2. Not recording time info in the `Lyric` class because the start time and the end time will be affected by the layout.
    - The start time and end time will not be saved in the layout metadata, or being calculated instead.
3. Re-define the roles:
    - Beatmap: record the supported layout type and it's info.
    - Hit object(Lyric): pure and most importand data, the data field should always useful and will not being replaced if display type changed.
    - Skin: give it an id, and it will return something like texture and shader effect.

andy840119:
The benefit of change:
- Missing skin in the beatmap will only lost the style. We can use internal or default skin instead.
    - Means less logic in the skin class.
- The data is not separated.
- Easy to design new layout. The current layout will not being affected if create new layout.
- Should be possible to switch the layout by mod.

andy840119:
Not really sure should use the same mapping role?
And the relationship between lyric and note?

andy840119:
What's the step to move the layout from skin into beatmap:
- [x] Remove initialize working time in the lyric -> Implemented in the #1763.
- [x] Implement classic `layout` class for the beatmap. -> Implemented in the #1821.
- [ ] Should be able to place the `layout` and `lyric config into` into hit-object by beatmap processor.
- [x] Make start time and end time as `JsonIgnored` -> Implemented in the #1761.
- [ ] Move the apply logic from karaoke skin to:
    1. Drawable class
    2. Layout in the beatmap.
- [ ] Remove the logic from the skin.

andy840119:
Not really sure should rename the "layout" into "stage"?
Because layout info might contain style lookup info.

andy840119:
Close this issue due to most parts has been implemented.
Lazy to mention all the PRs.

PR #1823, opened 2023-01-05
Make stage info more useful.

andy840119:
Add wrapping method in the stage info class for prevent create duplicated style or layout with same id.
Or add the wrong mapping roles.


andy840119:
Mark as relative of issue #1755.

Issue #1765, opened 2022-11-27
Implement the classic stage editor


andy840119:
Guess should implement the editor while thinking #1764 
And can start implement this shit after add the new layout class in the #1755.

prepare:
- [x] Classic layout modal implementation. -> Handle in the #1755.
- [ ] Design the figma page for it.
    - [ ] Not really sure will use the same screen for different layout? -> maybe just named classic layout editor screen.

andy840119:
![image](https://user-images.githubusercontent.com/9100368/204128623-72c68fb9-b530-4cfa-83cc-12e70e4b5b52.png)
Maybe what is this shit looks like.

andy840119:
诶不是
怎麼感覺沒有結束的一天

andy840119:
todo:
- [ ] Implement classic stage generator.
- [x] Implement classic stage change handler. -> Implemented in the #1825.
- [x] Implement the checks for the stage category. -> Implemented in the #1826.

andy840119:
Note:
Should implement the timing part for the #1828.

andy840119:
todo:
- [ ] Need to think about should combine the layout and timing into single editor?

PR #1826, opened 2023-01-07
Implement classic layout checks.

andy840119:
Implement part of #1765.
Implement the basic check for the classic stage.
Will add more check if needed.

What's done in this PR:
- Implement the base beatmap stage check. This check class will check the stage element category also.
- Implement the classic stage check.


andy840119:
Another 0.5K codes added.

PR #1836, opened 2023-01-13
Adjust timing info structure.

andy840119:
Adjust the structure implemented in the #1830
Add more mapping for record the relation between lyric id and the timing point id.

https://karaoke-dev.notion.site/Classic-stage-timing-info-05350c50c19d4d1fbccbeb5b9b0b2e07
Here's the reason why change the timing info structure.
As usual, written in Chinese.


andy840119:
Wait.
Or just remain that read/write methods but not change the format?
because very not sure create id is a good idea or not 🤔

andy840119:
I think it should be able to change the format back by another PR.
It should not that hard.

PR #1848, opened 2023-01-22
Implement auto generator change handler interface

andy840119:
Define the interface for the auto-generator.
Follow how `IEditorVerifier<T>` did in the #1841


andy840119:
I might be hard to define the common interface for the invalid messages.
But should be OK for now.

The most important part is letting developers able to find the auto-generator by the interface, or see how many change handler has auto-generate feature.

PR #1847, opened 2023-01-21
Upgrade package to the latest.

andy840119:
Upgrade some package to the latest to make the project happy.

PR #1850, opened 2023-01-23
Upgrade package to the latest

andy840119:
Upgrade the package to the latest again for fixing issue caused in the #1847


andy840119:
Guess should have no choice to merge it now because previous version of osu.game will cause all visual test case cannot be opened.

PR #1857, opened 2023-01-28
Implement compare utils for better to compare two objects.

andy840119:
In some cases, we need to compare object by some of the properties.
For example, if compare the `Tone.cs`, we need to compare the `Scale` first. If the scale in the tones are the same, than next is compare the `Half` property.
.
Should be better to write a utils for compare the object by list of the properties.


andy840119:
像極了前來進貢的卑微人類

PR #1871, opened 2023-03-07
Apply the setting source into the config class.

andy840119:
What's done in this PR:
- Remove the `CreateDefaultConfig()` method in the config interface because default value will be in the bindable property.
- Make the `IHasConfig` from interface into the base class.
- Use the bindable in the config class.
- Implement helper in the test case for able to clear all default value.

And for now, should still able to show the popover with config and able to edit the value.


andy840119:
嘔心瀝血改了那麼多，你跟我說只修改了不到三百行?

Issue #1869, opened 2023-03-05
Auto-generate control component for the auto-generate config.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/222934770-0e2fb21d-b5e6-4d10-87ec-323208c81d88.png)
![image](https://user-images.githubusercontent.com/9100368/222934772-4f3c5402-c501-48a8-bce9-872310628dc5.png)
Like the control in the mod and the skin editor, should be possible to auto-generate the control for the auto-generator and the auto-detector config.

And here's the step:
- [x] Remove the dialog. The dialog will not be used anymore. -> Implemented in the #1870.
- [x] Change the property in the config to the bindable. Also, should add the `SettingSource` attribute. -> Implemented in the #1871.
  - [x] Note that should create the `SettingCategory` attribute for able to generate the control with different category. Implemented in the #1871.
- [x] Use the auto-generate resource instead. -> Implemented in the #1873.
  - [x] Should create config's own `SettingSourceExtensions` because we want to generate the control  with `LabelledTextBox` -> Implemented in the #1872.
  - [ ] Add a test case to make sure that serializer/deserializer is working.

andy840119:
Need to think about will only this part use the `SettingCategory ` in the editor?
Should make this more generic if will be used in the other part.

andy840119:
Note the benefit is use the bindable:
- It might be easy to roll-back to the default value by just a click.
<img width="519" alt="image" src="https://user-images.githubusercontent.com/9100368/223029222-12f8aa94-6a23-41c6-b291-7379292dd95e.png">

Also:
- [x] Think about should assign the default value into the property or `CreateDefaultValue()` interface? We might need the config instance that all value should be initialize(like false, zero or something). -> Guess answer is in the #1871 ?

andy840119:
Guess this issue can be closed now because most of issues has been implemented.
Serializer/deserialize issue can be fixed after.

PR #1873, opened 2023-03-11
Implement popover for able to auto-generate the control for the config.

andy840119:
Guess should be the main part of #1869

What's done in this PR:
- Adjust the category in the config.
- Should be able to get the generator config class without assign the type in the config manager.
- Implement new popover for able to auto-generate the control for the config.
- It's time to remove the old popover.
- Move some components into more generic location.


andy840119:
Nice 
Seems saved around 600 lines of code.

PR #1874, opened 2023-03-11
Should be able to get generator config just by type.

andy840119:
Inspired while implementing the #1873.


andy840119:
Just mark as related to the #1869.

PR #1876, opened 2023-03-19
framework upgrades

WebFreak001:
depends on https://github.com/karaoke-dev/osu-framework-font/pull/321 and needs new tag there + update in csproj which I haven't done here.


andy840119:
Package has been updated in the #1887.

PR #1879, opened 2023-03-22
Implement hit-object validator to check which property is invalid(need to update) in the beatmap processor.

andy840119:
What's done in this PR:
- Implement the hit-object validator. This validator can record which property is invalid.
- Use this validator in the note and the lyric.


andy840119:
todo:
- [ ] Think about is there any better naming for it? like `sync` or `process value`?

andy840119:
Not figure out better naming so just merge 😢

PR #1887, opened 2023-03-26
Update package to the latest.

andy840119:
Update package to the latest.
Seems workable in the DirectX11, nice.


andy840119:
Got no idea why the test case broken in here.
So just make the tag release first.

PR #1890, opened 2023-03-27
Implement generic interface to invalidate the property.

andy840119:
Implement the interface for the hit-object with working property has two reason:
1. Making the invalidate state cannot be modify easily.
2. Easy to invalidate the working property in the hit-object by the interface.


andy840119:
image
Damn need 6 minutes to run the full tests now 😢

PR #1894, opened 2023-03-29
Should validate the property as default.

andy840119:
Initial working property sync state should be valid.
Because there's not reason to mark the property as invalid if data property and working property are all empty.


andy840119:
image
Damn, both of them not working.

PR #1919, opened 2023-04-07
Add stage element as working property.

andy840119:
What's done in this PR:
- Add the stage elements as working property in the note and lyric.

The property will be used to:
1. let the stage editor know the element mapping result.
2. in the gameplay, will use the elements and the stage info to generate the layout, style or animation effect.(maybe get the stage info from karaoke beatmap by injection or another working property i think?)

### DIFF
--- a/osu.Game.Rulesets.Karaoke.Tests/Objects/Workings/LyricWorkingPropertyValidatorTest.cs
+++ b/osu.Game.Rulesets.Karaoke.Tests/Objects/Workings/LyricWorkingPropertyValidatorTest.cs
@@ -4,7 +4,9 @@
 using System;
 using J2N.Collections.Generic;
 using NUnit.Framework;
+using osu.Framework.Bindables;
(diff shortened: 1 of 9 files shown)


andy840119:
I guess should be better to save the stage info as working property in the lyric also 🤔 

The benefit:
1. There's no need to use the injection to get the stage info in the drawable lyric or note.
2. Easier for testing. We can just create one lyric then put it into the playfield.

andy840119:
Wait.
Need to think about that because we should be able to get the `StageElements` from the stage info by lyric, so technically there's no need to process it in the beatmap processor.
Guess the better way is saving the definition class only(e.g. `ClassicLyricLayoutDefinition`) in the lyric?

andy840119:
I think should be better to create something like `ElementProcessor` and save as working property in the lyric?
Will have element processor for the note and the lyric.

Until processor haven't finished or being defined, i think it should be OK to give it a merge.

PR #1922, opened 2023-04-07
Use the `current stage info` property in the beatmap instead.

andy840119:
For now, `CurrentStageInfo` is the final stage info that working in the gameplay and the editor.

What's done in this PR:
1. Fill the current stage info in the `beatmap processor`, should get the current stage from the karaoke beatmap directly in the gameplay or in the editor.
2. Use the `current stage info` in the change handler.
3. Should mark the `current stage info` as empty if remove the stage info.


andy840119:
Damn, the CI is not happy now 😢

Issue #1920, opened 2023-04-07
Implement stage effect applier for the note and the lyric.


andy840119:
Continuous of issue #1919.

Guess should create the processor to convert the stage element(s) into the thansformer.
And the will create the stage processor for the lyric and the note.

andy840119:
The next step is implement algorithm for calculate the lyric moving timing point(#1906)
And think about how to apply the effect(#1920)

andy840119:
todo:
- [ ] Actually apply the applier to the note and lyric. 
- [ ] Actually apply the applier to the playfied.
- [ ] Should create a case for the #1914 because translate/singer position might not always follow the `lyric`.

PR #1930, opened 2023-04-13
Implement stage effect applier for the note and the lyric.

andy840119:
Implement the model part of the #1920.

What's done in this PR:
1. define the applier for the lyric and the note.
2. use the applier in the lyric and the note.


andy840119:
Not really sure will change the architecture before implement the animation layout.
But it's enough for the classic and the default stage info.

andy840119:
幹這個 CI 越來越慢了QQ

PR #1938, opened 2023-04-16
Implement the playfield stage applier.

andy840119:
Implement part of #1920.
Will use this stage applier for able to change the arrangement if using different stage.


andy840119:
Note:
the benefit to use the `TransformSequence<LyricPlayfield> transformSequence` is because we might able to roll-back the property in the lyric playfied like:
```csharp
protected override void UpdateLyricPlayfieldArrangement(TransformSequence<LyricPlayfield> transformSequence, bool displayNotePlayfield)
{
    transformSequence.TransformTo(nameof(LyricPlayfield.RelativePositionAxes), Axes.X);
}
```

PR #1955, opened 2023-04-23
Implement classic stage applier.

andy840119:
image
Seems able to make the PR because most of the part in the hit-object/playfield stage applier has been implemented.

What's done in this PR:
- Adjust some small part in the classic layout/timing auto-generator.
- Add more params in the classic stage definition.
- Implement the hit-object applier.
- Implement the playfield applier.


andy840119:
Guess should still give it a merge because local test passed.

PR #1956, opened 2023-04-24
Implement the stage info generator selector.

andy840119:
Implement the selector for able to select the stage info want to generate.


andy840119:
Guess we should rename some methods in the `IBeatmapStagesChangeHandler` to make it more generic.
Maybe should inherit the `IAutoGenerateChangeHandler` because the stage is being generated also?

Issue #1957, opened 2023-04-24
Separate the auto-generate logic in the LyricAutoGenerateChangeHandler.


andy840119:
I think it might not be a good way to collect all the generators into single change handler.

Guess should remove the `LyricAutoGenerateChangeHandler` eventually.

andy840119:
I think `AutoGenerateSubsection` should include the generator inherit the `IAutoGenerateChangeHandler`?
But it will not able to call the change handler if inherit other kind of `IAutoGenerateChangeHandler` (e.g. `IAutoGenerateChangeHandler<T>`)

Need more think about this.

andy840119:
Implemented, nice.

andy840119:
It might not be a good idea if one property has more than 2 generators.

In #2004, time-tags change handler has 2 generator
- generate the time-tags.
- generate the romaji inside the time-tags.

Guess should be better to collect all generators again.

PR #1958, opened 2023-04-24
Separate the auto-generate logic in the LyricAutoGenerateChangeHandler.

andy840119:
Closes issue #1957.
As we are already use one change handler to edit one property in the lyric, we should move the auto-generator logics into those change handlers.


andy840119:
Use one change handler to edit one property does not means one property only has exactly one generator Orz.

PR #1960, opened 2023-04-26
Should invlidate the working property in the mod because the stage will apply to the hit-object while selecting them in the song selection.

andy840119:
What's done in this issue:
1. Change the mod should invalidate all the working property that is related to the stage info.
2. Adjust some properties in the stage mod.
3. Add the mod into the karaoke ruleset.

### DIFF
--- a/osu.Game.Rulesets.Karaoke/Mods/ModStage.cs
+++ b/osu.Game.Rulesets.Karaoke/Mods/ModStage.cs
@@ -3,9 +3,12 @@
 
 using System;
 using System.Linq;
+using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Game.Beatmaps;
 using osu.Game.Rulesets.Karaoke.Beatmaps;
 using osu.Game.Rulesets.Karaoke.Beatmaps.Stages;
+using osu.Game.Rulesets.Karaoke.Objects;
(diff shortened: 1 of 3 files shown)


osu.Game.Rulesets.Karaoke/Mods/ModStage.cs
```diff
@@ -37,6 +33,15 @@ public void ApplyToBeatmap(IBeatmap beatmap)
...
         karaokeBeatmap.CurrentStageInfo = matchedStageInfo ?? CreateStageInfo(karaokeBeatmap)!;
+
+        // should invalidate the working property here because the stage info is changed.
+        // has the same logic in the beatmap processor.
+        beatmap.HitObjects.OfType<Lyric>().ForEach(x =>
+        {
+            x.InvalidateWorkingProperty(LyricWorkingProperty.Timing);
+            x.InvalidateWorkingProperty(LyricWorkingProperty.EffectApplier);
+        });
+        beatmap.HitObjects.OfType<Note>().ForEach(x => x.InvalidateWorkingProperty(NoteWorkingProperty.EffectApplier));
```
andy840119, line 44: Remove this code if make the stage into individual namespace and give it something like `stage processor`.
We will move all the stage invalidate logic into there.

PR #1961, opened 2023-04-26
Remove the support for able to convert the `.nkmproj` (Nicokara file extension) into the skin.

andy840119:
Because the `nicokara format` will have lot's of data loss if convert to the skin/classic stage ;_;


andy840119:
image
Damn the .nkmproj file is fat 😂

PR #1967, opened 2023-04-30
Updating main project to use File Scoped Namespaces.

andy840119:
image
Last part of #1754


andy840119:
Damn, i forgot will trigger the CD.

Issue #1971, opened 2023-05-02
Should be able to click the error info to show all the invalid rows.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/235728440-db9298d7-5d95-48cf-b642-f23f6f7a9329.png)
After click the error in here.

![image](https://user-images.githubusercontent.com/9100368/235728500-8de52945-7726-4281-9ca3-d8c418302994.png)
Should be useful if able to show the table like how edit time-tag mode did in the popover.
User can click to navigate to the target edit mode with just a click.

andy840119:
Implemented.

PR #1975, opened 2023-05-03
Should be able to click the error info to show all the invalid rows.

andy840119:
image
Closes issue #1971.


andy840119:
Not very sure how to show the scrollbar if has too much rows.
But it's OK for now.

PR #1979, opened 2023-05-05
Add the damn separators into the toolbar.

andy840119:
image
What's done in this PR:
- Add the separator between moving buttons and the edit buttons
- Add the moving buttons for the texting mode.
- Create the easier way to get the SubMode with target type.


andy840119:
總覺得 CI 速度變快了?

PR #1980, opened 2023-05-06
Use utils to get the value by state.

andy840119:
Should use:
```csharp
var value = TextIndexUtils.GetValueByState(textIndex, startValue, endValue);
```

to replace:
```csharp
var value =  state switch
{
    TextIndex.IndexState.Start => startValue,
    TextIndex.IndexState.End => endValue,
    _ => throw new ArgumentOutOfRangeException(nameof(state))
};
```


andy840119:
maybe we should add the alert if use switch with `TextIndex.IndexState`?

PR #1981, opened 2023-05-06
Refactor caret position provide method

andy840119:
What's done in this PR:
1. Add the test case for the `InteractableKaraokeSpriteText`.
2. Test case is able to show what's get by the position and see the element's rect or position.
3. Re-write the method for able to get the element or element position by different type(see: https://karaoke-dev.notion.site/Refactor-method-in-the-InteractableKaraokeSpriteText-35b05a5a55b64e5cbf5ae30a075f1917)
4. Remove some old interfaces and use the new one.

image
image
And it's the demo that we can get the rect by character or indicator(spacing between characters).


andy840119:
這次就先讓你 merge 了
下不為例

Issue #1973, opened 2023-05-02
Re-design the caret algorithm for the create/remove time-tag mode.


andy840119:
![image](https://user-images.githubusercontent.com/9100368/235729295-bb2ec2a4-c103-4e39-9a79-0d160cff7b7d.png)
In the create time-tag mode, we use the `TimeTagIndexCaretPositionAlgorithm` to move the cursor.
But it might be slow 

![image](https://user-images.githubusercontent.com/9100368/235730198-d75e799b-10e6-4a82-a33e-f4b9e9b7fdca.png)
It might be better if able to focus to the whole character.

![image](https://user-images.githubusercontent.com/9100368/235731763-56c53f44-bc4f-41bf-8951-46aa70392409.png)
In here, we should add 4 keys for the create/remove time-tag.
- Create start time-tag.
- Create end time-tag.
- Remove start time-tag
- Remove end time-tag.

andy840119:
Also, should think able is there any need to modify the time-tag index?
Guess it only used to adjust the import lyric?

andy840119:
Not very sure should have click area like how control point editor did?
![image](https://user-images.githubusercontent.com/9100368/235732853-5f05ab67-e824-4083-915e-195536b888f1.png)

[Update]
This topic has been moved to the #1974.

andy840119:
Use the new caret has been implemented in the #1984

PR #1984, opened 2023-05-07
Redesign caret algorithm for the create remove time tag

andy840119:
image
Implement the main part of the issue #1973.

What's done in this PR:
- Implement the new caret position algorithm.
- Adjust the key binding. For now, we have 2 keys for add the start/end time-tag and another 2 is for delete them.
- It's time to say goodbye to the old drawable caret and algorithm.


andy840119:
Let this PR merge first.
Then will take times to think about how to adjust the modify mode.

Issue #2020, opened 2023-06-02
Combine the `Validate` and `PositionMovable` in the `CaretPositionAlgorithm`


andy840119:
Technically, Validate can just check the result in the PositionMovable?

todo:
need to check the `TimeTagCaretPositionAlgorithm`.

andy840119:
Implemented in the 2021.

PR #2021, opened 2023-06-03
Adjust caret position usage.

andy840119:
What's done in this PR:
- Algorithm should not accept the null caret position index.
- There's no need to implement the pre-validator unless need to make the check not that strict(#2020).
- Separate the get caret position index logic from the `InteractableLyric` because drag to change the caret index might use the same logic.


andy840119:
算了一下，這個星星真是有夠難賺

Issue #2040, opened 2023-06-25
Use Grid instead of int as primary key.


andy840119:
Instead of finding the largest index from the list of hit-object, like:
- SingerInfo.getNewSingerId();
- LyricsChangeHandler.getId();

maybe should be better to just use the `GUID` to prevent the duplicated id.

Another benefit:
- There's no need to deal with logic to re-sort the id([1,3,4] -> [1,2,3])
- Technically, there's no possible to have duplicated id.
- Able to generate the id inside the class, follow how `SkinInfo` did in the osu.game
- Easily to global search the id in the json file.

And note that:
- We can `change the algorithm for make the unique id shorter or not easier to duplicated` at any time.

andy840119:
https://dvoituron.com/2022/04/07/generate-small-unique-identifier/
Seems GUID is too large, maybe should find another way to shorter the unique id's length.

andy840119:
Those objects need the id:

Beatmap:
- Singer -> max 20 singers
- Lyric -> max 300 lyrics

Stage:
- StageElement(i'm not very sure if skin element really need to unique key) -> max 200 elements
- ClassicLyricTimingPoint(not really sure also.) -> max 2000 timing point.

Seems duplicated id is not the main issue.

andy840119:
I think 7 characters as primary key is enough.
like how git generate the SHA1?

andy840119:
And we need to make sure that the generated singer or lyric's id should not be null in the test case.

andy840119:
Not very sure if it's over-design but we can follow how `ResolvedAttribute` did to assign the value to the primary key.

andy840119:
Or maybe make the id has it's own struct?
E.g. `TinyUuid` or `NanoUuid`
see: https://stackoverflow.com/a/61777945/4105113

Or `ShortId`
see: https://stackoverflow.com/a/45987192/4105113

andy840119:
Note:
Maybe Guid should accept the string format, so there will be no need to write it's own serializer.

See:
https://github.com/dotnet/runtime/issues/31465

andy840119:
There's the sample code how serialize/deserialize the Guid without customized serializer.
see:
```
[Test]
public static void GuidSerializationTest()
{
    string str = JsonConvert.SerializeObject(new Foo
    {
        MyGuid = Guid.NewGuid()
    });
    Assert.NotNull(str, str);
}

[Test]
public static void GuidDeserializationTest()
{
[...]

PR #2045, opened 2023-07-02
Create element id struct.

andy840119:
First implementation of #2040.
Create the struct like `Guid` to define karaoke ruleset's own id format.

The format idea is from the git commit sha, the main goal is:
1. Easy to create the id. 
    - It's easy to make the mistake if need to get all related objects if want to create the new id.
    - Able to generate the id everytime object created.
2. No need to write the algorithm to refresh(or called re-calculate) the ids(e.g. [1,4,5] -> [1,2,3]). 
3. (not very important) Id should not include the information about create order.

and we should avoid:
1. the id is not easy to conflict.

The algorithm in the `ElementId.NewElementId()` can be changed at any time, just need to make sure that the id should be 7 digits and within `0~9` or `a~f`.


andy840119:
Quick-peek:
image
The id like `Singer ids` stored inside the `lyric` will be stored like this.

PR #2046, opened 2023-07-02
Change the singer id type from int to the customized format.

andy840119:
Part of #2040.


andy840119:
I think should be OK to give it a merge.

PR #2047, opened 2023-07-03
Singer extension should be able to accept the int number directly.

andy840119:
Improvement part for the #2046.
It's OK because this extension is for testing propose only, so make the test easier is the first priority.


andy840119:
Just mark as related to the #2040.

Issue #2048, opened 2023-07-04
Able to write the value into the property with protected/private setter.


andy840119:
Found while implementing the #2040

It need to manually deal with json serializer.
But should auto-add it eventually.

andy840119:
Note:
Can just add the `[JsonProperty]` in the id property.

andy840119:
Fixed.

Issue #2054, opened 2023-07-04
Should change the id after `DeepClone()` the lyric.


andy840119:
Found in the #2040 also.

andy840119:
Fixed.
Also provide the extension for able to change the primary key. this utils should not be called often technically.

PR #2055, opened 2023-07-04
Change lyric id format.

andy840119:
Implement part of #2040.
And issue #2048 is avoided in this PR.


andy840119:
Bug #2054 is noticed in this PR.
It's OK to give it a pass.
Will fix the remaining issue in the next PR (maybe tomorrow).

PR #2057, opened 2023-07-05
Should change the id after `DeepClone()` the lyric.

andy840119:
Improve the code quality and resole issue #2054 .


andy840119:
Not very sure will it be better if only let user can random change the id in the `PrimaryKeyObjectExtension`.
But it's OK for now.

PR #2060, opened 2023-07-09
Able to get all matched element ids

andy840119:
For able to testing the primary key in the element category (see: #2040)


andy840119:
Mark as needed in the #2061

Issue #991, opened 2021-12-29
Normalise trailing commas in code


andy840119:
Make a change until https://github.com/ppy/osu/pull/16275/files merged.

andy840119:
Almost a year.

andy840119:
Another 0.5 years

PR #2066, opened 2023-07-15
Update dotnet setting from osu project

andy840119:
Preparation of issue #991.
Need to update the `DotSettings.xml` from the `osu.game` first.


andy840119:
Can give it a merge without CI check because did not touch the code.

Issue #502, opened 2021-03-01
Filter by tag regex


andy840119:
See line 13 in  `FilterQueryParser`
```
private static readonly Regex query_syntax_regex = new Regex(
            @"\b(?<key>stars|ar|dr|hp|cs|divisor|length|objects|bpm|status|creator|artist)(?<op>[=:><]+)(?<value>("".*"")|(\S*))",
            RegexOptions.Compiled | RegexOptions.IgnoreCase);
```

And here's the example(https://regex101.com/r/2BvxFu/1).
Not really sure will be use in this project.
Might be use in test case for fast generate target test object.

andy840119:
This document issue can be closed due to possible to find this issue easily by chatGPT(?)

Issue #2004, opened 2023-05-12
Move the romaji to the time-tag.


andy840119:
Not very sure about that but it will be useful to show the romaji-only lyric if we make every char has it's own romaji.

andy840119:
我又懶了，用一下中文好了

![image](https://github.com/karaoke-dev/karaoke/assets/9100368/08952de6-3b7f-4981-beb6-3fc8caa238d3)
目前 romaji 的設計是歌詞的一小段會
而不是每一個字元都有對應的 romaji
如果有一天需要做出只會顯示 romaji 的歌詞，會有 time-tag 對不上的問題

Recorded in here:
https://karaoke-dev.notion.site/tomaji-text-tag-time-tag-8c351278062a470fa4449a9172b3ad02?pvs=4

andy840119:
另外，或許需要設計更好的 tool，去 combine and separate the romaji?

andy840119:
Guess:
1. Remove the romaji text class and add the romaji property into the time-tag.
2. Add the `RomajiContinuous` property in the time-tag.

想法是一個 time-tag 通常對應一個音節，romaji也有類似的分段方式

andy840119:
https://randulakoralage82.medium.com/format-your-net-code-with-git-hooks-a0dc33f68048
here's the article about how to set-up the pre-commit in the C# project.

andy840119:
Remaining:
- [x] Tag helper should be able to support the text within the time-tag. -> Implemented in the #2076
- [ ] Should be able to support save/load the text in the time-tag serializer. -> Moved into #2278
- [x] Implement the auto-generator for able to fill the romaji to the time-tag. -> Implemented in the #2077
- [x] Should be able to edit the romaji text and the initial romaji property in the change handler. -> Implemented in the #2079
- [x] Add time-tag romaji auto-generate logic for the time-tag change handler. -> Implemented in the #2004.

andy840119:
Wait.
Note has `Ruby text` property.
Not very sure will have `Romaji` property in the note also? -> guess should be get from time-tag, or note should have time-tag index?

andy840119:
Remaining part(editor):
- [x] Should edit the property from the time-tag, not the romaji class. #2136
- [x] Adjust the import step. Auto-generate romaji should be after generate time-tag. Or just remove this step. #2137
- [x] Remove the old romaji generator(and the generator). #2138
- [x] Remove the change handler. #2139
[...]

PR #2073, opened 2023-07-21
Refactor the test case helper.

andy840119:
It's a good start to let the tag helper more flexibility and easy to add more feature.

The final goal might like #502
Will be easy to add new properties for the tag in the test tag helper.


andy840119:
天啊
那個CI應該處理一下了QQ

andy840119:
Mark as preparation for the #2004.

PR #2082, opened 2023-07-29
Remove generate type property in the caret position.

andy840119:
The reason to save the `CaretGenerateType` is because need to check the caret is moved by mouse or the keyboard.
But have better solution to deal with that, so remove this property from the caret position.


andy840119:
Also:
```csharp
public enum CaretGenerateType
{
    // OK we know that caret might move by action (e.g. keyboard)
    Action,

    // But we cannot know that the lyric is moved by mouse, and it might be hard to know why can't select the first item to the blueprint if has this value in the caret object.
    TargetLyric,
}
```

andy840119:
image
Man i should disable this shit until fixed 😢

PR #2089, opened 2023-08-06
Move stage into individual namespace.

andy840119:
Stage should not be the part of metadata in the beatmap eventually.


andy840119:
Mark as related to the #2090

PR #2098, opened 2023-08-15
Adjust the caret state behavior.

andy840119:
What's done in this PR:
- Adjust some method order in the lyric editor caret state.
- Handle the action (e.g. move to previous index) if has range caret position. Should follow the same behavior as text editor in the windows/osx.


andy840119:
Fuck that CI

PR #2105, opened 2023-08-27
Remove `BeatmapEditorRoundedScreen`

### LINKED ISSUE #1341: Remove KaraokeEditorRoundedScreen screen.
image

After new design of editor(#1340), the rounded screen style has been removed.
So should follow the change to adjust the singer editor and translate editor eventually.

But might not that urgent because it's still OK for now.

Joehuu:
- Resolves https://github.com/karaoke-dev/karaoke/issues/1341


andy840119:
image

I guess i should apprvoe this PR until official editor is more complete.
Because some of the editor looks weird if display in the full-screen. It's hard for me to point-out the reason and find a way to adjust that 😢

andy840119:
And if you are interested in contributing to this project, i can filter the issue and pick-up the suitable one for you(because some of the issues is a memo for me and might be hard to see the priority)

Joehuu:
> Because some of the editor looks weird if display in the full-screen. It's hard for me to point-out the reason and find a way to adjust that 😢

The change is to mostly match the current design state of the official editor. I do agree that some screens look weird because they are mostly blank / have no content yet (even in the official editor), and making it fullscreen (and hiding the beatmap background) doesn't make it better.

Can wait, it's just a cosmetic change.

> And if you are interested in contributing to this project, i can filter the issue and pick-up the suitable one for you(because some of the issues is a memo for me and might be hard to see the priority)

I just want to clean up the issues right now by targeting the easier ones first (even if it's low priority). That would be `code quality`, `design`, `bug`/crashes, etc.

I don't want to be recommended an issue up front, but I would appreciate `priority:1|2|3` or similar labels like osu! so I can get a sense of what the current goals are and contribute to something bigger when I get the chance.

andy840119:
> I don't want to be recommended an issue up front, but I would appreciate priority:1|2|3 or similar labels like osu! so I can get a sense of what the current goals are and contribute to something bigger when I get the chance.

I'm not sure using `priority` flag is a good idea because:
1. the main contributor for this project is only me ;_;
2. this project does not have much users right now(means there's no s1 issue), so priority is decided by what i'm doing right now and might be changed frequently.

I guess using  https://github.com/karaoke-dev/karaoke/labels/Good%20for%20contributor flag might be better right now, you can pick-up the issue with this label, and feel free to ask me if have any question.

Issue #2007, opened 2023-05-13
Should be able to drag the main lyric text to generate the ruby/romaji text.


andy840119:
<img width="693" alt="image" src="https://github.com/karaoke-dev/karaoke/assets/9100368/2aee131e-1d1a-4f24-b23d-7a039221ad2a">

As title, instead of typing the start and end index in here, should be able to drag the the main text to create the ruby/romaji with start and end index.

andy840119:
Guess can implement after #2008?

andy840119:
Guess should wait after #2009?

[Update]
Implemented

andy840119:
After we have sample to drag range of the lyric and do the action (see: #2099)
Maybe it's a good start to continue this issue.

andy840119:
Remaining part:
- [x] Guess should be better to have the pop-up `with ruby text input and create button` after select the range of the text. #2126
  - [x] caret state should have bindable for able to record the release state. #2125
- [x] Need to show the popover if click the ruby tag, like how note edit area did. #2131
- [x] Create a hot-key for able to create ruby-tag if select the range. -> will open the popover after click or select range of the main text.
  - E.g. create the ruby tag if select the range of the text and press "enter".
  - Also, should show the description by edit mode switcher subsection.
- [x] It's time to remove the "add ruby" and "add romaji" button. #2128

andy840119:
Guess it's enough to close this issue.

PR #2125, opened 2023-09-23
Change the drag caret position will not change the caret position.

andy840119:
As title.

Need to make sure that caret position will have value if click or drag the lyric before.
But should be better to make them as individual behavior.

Means "caret position" and "drag caret position" should not exists at the same time.

Also.
Expose the caret position algorithm for easier to let outside know what is current algorithm is using.
So they can prepare the caret index position or create the drawable caret.

### DIFF
--- a/osu.Game.Rulesets.Karaoke.Tests/Screens/Edit/Beatmap/Lyrics/States/LyricCaretStateSwitchModeTest.cs
+++ b/osu.Game.Rulesets.Karaoke.Tests/Screens/Edit/Beatmap/Lyrics/States/LyricCaretStateSwitchModeTest.cs
@@ -141,13 +141,13 @@ public partial class LyricCaretStateSwitchModeTest : BaseLyricCaretStateTest
 
     protected void AssertCaretPositionAlgorithmIsNull()
     {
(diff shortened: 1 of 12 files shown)


andy840119:
Guess should mark as related to the #2007.

osu.Game.Rulesets.Karaoke.Tests/Screens/Edit/Beatmap/Lyrics/States/LyricCaretStateSwitchModeTest.cs
```diff
@@ -130,5 +139,16 @@ protected void AssertCaretDraggable(bool caretDraggable)
         AddAssert("Assert caret draggable", () => LyricCaretState.CaretDraggable == caretDraggable);
     }
 
+    protected void AssertCaretPositionAlgorithmIsNull()
+    {
+        AddAssert("Assert caret position type", () => LyricCaretState.CaretPositionAlgorithm == null);
```
andy840119: "Assert caret position algorithm should be null."

andy840119 (commented):
Other part LGTM.

Issue #2141, opened 2023-10-01
Startup Error


Sewlell:
```
2023-10-01 04:43:06 [verbose]: ----------------------------------------------------------
2023-10-01 04:43:06 [verbose]: runtime Log for user name (LogLevel: Verbose)
2023-10-01 04:43:06 [verbose]: Running osu 2023.924.1.0 on .NET 6.0.22
2023-10-01 04:43:06 [verbose]: Environment: Windows (Microsoft Windows NT 10.0.22621.0), 8 cores
2023-10-01 04:43:06 [verbose]: ----------------------------------------------------------
2023-10-01 04:43:06 [verbose]: 🖼️ Configuration renderer choice: Automatic
2023-10-01 04:43:06 [verbose]: 🖼️ Renderer fallback order: [ Direct3D 11, OpenGL (Legacy), OpenGL, Vulkan ]
2023-10-01 04:43:06 [verbose]: 🖼️ Initialising "Veldrid" renderer with "Direct3D11" surface
2023-10-01 04:43:06 [verbose]: Updated display mode to desktop resolution: 1920x1080@60, SDL_PIXELFORMAT_RGB888
2023-10-01 04:43:07 [verbose]: Direct3D 11 Initialized
2023-10-01 04:43:07 [verbose]: Direct3D 11 Feature Level:           11.0
2023-10-01 04:43:07 [verbose]: Direct3D 11 Adapter:                 NVIDIA GeForce MX130
2023-10-01 04:43:07 [verbose]: Direct3D 11 Dedicated Video Memory:  2008 MB
2023-10-01 04:43:07 [verbose]: Direct3D 11 Dedicated System Memory: 0 MB
2023-10-01 04:43:07 [verbose]: Direct3D 11 Shared System Memory:    6045 MB
2023-10-01 04:43:07 [verbose]: UseStructuredBuffers: True
2023-10-01 04:43:07 [verbose]: 🖼️ Renderer initialised!
2023-10-01 04:43:07 [verbose]: Host execution state changed to Running
2023-10-01 04:43:07 [verbose]: Execution mode changed to MultiThreaded
2023-10-01 04:43:07 [verbose]: Beginning realm file store cleanup
2023-10-01 04:43:07 [verbose]: Finished realm file store cleanup (0 of 20809 deleted)
2023-10-01 04:43:07 [verbose]: 🔈 BASS initialised
2023-10-01 04:43:07 [verbose]: BASS version:           2.4.16.7
[...]

PR #2142, opened 2023-10-01
Fix invalid culture code cause cannot load the ruleset.

andy840119:
Fix issue #2141.
Instead of throw the exception if the culture code is not valid, should reset it and log it instead.

todo:
- [x] Check issue fixed.


andy840119:
Seems this issue fixed 👍🏼

PR #2165, opened 2024-01-16
Rename from "romaji" into "romanization" in the time-tag.

andy840119:
Part of #2145

Those parts are renamed:
- Time-tag.
- Change handler.
- Generator.


andy840119:
Related to #2145.

PR #2180, opened 2024-03-03
Romanization -> Romanisation

andy840119:
Change the English wording to UK wording.
Like lazer use `colour` instead of `color`.


andy840119:
Change the wording because Lazer already using it for those non-native song name.

see:
https://github.com/ppy/osu/pull/16273/files

andy840119:
Related to #2145.

PR #2196, opened 2024-03-17
Rename remaining `romaji` to `romanisation`.

andy840119:
As title.
Last part of #2145.


andy840119:
image
All test passed in local.

PR #2232, opened 2024-05-18
Add the architecture test for this project.

andy840119:
Part of #2213.
As project become larger and larger, it's hard to create/modify code with correct coding style/architecture by memory.
And the decision detail might be forgot if not touch the module for several months. 
It's time to make the memory into automation script to check the error.

What's have in this project:
- Write the architecture decision or architecture decision record(ARD), see: https://github.com/joelparkerhenderson/architecture-decision-record
- Actually check the architecture by code.


andy840119:
- [x] Appveyor should actually test new project.
- [x] Test this project in the osx or linux environment.

andy840119:
image
Console says it runs the test, but the test is not listed in the test result 🥲 
Guess it should be testsed 🤔

PR #2307, opened 2024-12-06
Remove layout definition from skin.

andy840119:
As title.


andy840119:
Have no idea why appveyor says no, but test is runnable in the local.
Maybe it's time to change to the github action 😢

PR #2316, opened 2025-01-20
Add Metadata and Resources sections to the editor's setup screen

CloneWith:
Metadata and resource editing are essential features for a beatmap, and adding them should be fine.
Also changing the items' type to `Drawable` so it's convenient to "group" sections together with `FillFlowContainer` or something else X)

### DIFF
--- a/osu.Game.Rulesets.Karaoke/KaraokeRuleset.cs
+++ b/osu.Game.Rulesets.Karaoke/KaraokeRuleset.cs
@@ -284,12 +284,14 @@ public partial class KaraokeRuleset : Ruleset
         return statistic.ToArray();
     }
 
-    public override IEnumerable<SetupSection> CreateEditorSetupSections() => new SetupSection[]
-    {
+    public override IEnumerable<Drawable> CreateEditorSetupSections() =>
+    [
+        new MetadataSection(),
+        new ResourcesSection(),
         new KaraokeSingerSection(),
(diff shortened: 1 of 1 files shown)


andy840119:
https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Osu/OsuRuleset.cs#L343-L363
https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Taiko/TaikoRuleset.cs#L195-L198

Seems other rulesets add this section also.

osu.Game.Rulesets.Karaoke/KaraokeRuleset.cs
```diff
@@ -284,12 +284,14 @@ public override StatisticItem[] CreateStatisticsForScore(ScoreInfo score, IBeatm
...
-    {
+    public override IEnumerable<Drawable> CreateEditorSetupSections() =>
+    [
+        new MetadataSection(),
+        new ResourcesSection(),
         new KaraokeSingerSection(),
         new KaraokeTranslationSection(),
         new KaraokeNoteSection(),
-    };
+    ];
```
andy840119: Note:
Still not decide using the latest net language for the array, so please change back to `[]{ ... }`

andy840119 (approved):
Overall LG™
Please fix the issue by suggestion.
And the CI broken can ignore now.

PR #2327, opened 2025-06-20
Mark Classic and Preview Stage mods incompatible with each other

CloneWith:
*This is a quick bug fix.*

In the latest release of karaoke!, it's possible to select these two mods at once, and such combination makes it failed to start the gameplay.


andy840119 (approved):
Maybe the IncompatibleMods  should be something like  `IModStage`, but there's no such mode now.

CloneWith:
> Maybe the IncompatibleMods should be something like `IModStage`, but there's no such mode now.

Can be useful when we have more stage designs.

andy840119:
Indeed.

PR #2329, opened 2025-06-21
Use NUnit 4.x for tests

### LINKED ISSUE #2150: Upgrade NUnit to 4.0
image

Will have lots of api breaking change after upgrade to nunit 4.0
Some methods moved to LegacyAssert.

see: 
https://docs.nunit.org/articles/nunit/release-notes/breaking-changes.html#nunit-40

Before start upgrade the nunit, maybe should finish at least one:
- [ ] Wait until osu project upgrade first (?)
- [ ] Understand why nunit make such changes.

CloneWith:
Resolves #2150.

Quite large amount of repetitive work, which won't be completed efficiently without AI agent's help. 😅

In my development environment `KaraokeDifficultyCalculatorTest` raised `MissingMethodException`. This is worth investigating.

*Update*: This is because ppy's project is still using Nunit 3.x.


andy840119:
> In my development environment KaraokeDifficultyCalculatorTest raised MissingMethodException

Yes, o!f and lazer still using 3.x
But [some of the project](https://github.com/karaoke-dev/osu-framework-microphone/blob/master/osu.Framework.Microphone.Tests/osu.Framework.Microphone.Tests.csproj#L16) can upgrade the nUnit to 4.x and i need to check why only this test case broken 😢

andy840119:
Most tests passed as before.
Thanks for the contribution ❤️

PR #2331, opened 2025-07-26
Hide default left sidebar in ruleset settings

CloneWith:
This is done by hiding sidebar *content* in `PopIn()`. Also needed to reset the margin of the content container.


andy840119:
LGTM for me if it works.

PR #2332, opened 2025-08-13
Minor adjustments of settings panel UI

CloneWith:
## Changes

- Behaviour of hover background highlight
- Layering change of preview area


andy840119:
Nice

PR #2333, opened 2025-08-27
Rewrite logic of getting overlay placement container

CloneWith:
The latest version of this ruleset won't display the changelog overlay properly. It would appear in the FPS counter display at the bottom right corner... So I make this PR to attempt to resolve this issue.

For stability and performance concerns, it should be better to use `WaveOverlayContainer` to locate our target. Below is what I got when debugging for reference:

图片


andy840119:
LGTM for me if it works 🚀

PR #2342, opened 2026-08-13
Embed karaoke resources in the ruleset assembly

### LINKED ISSUE #2337: ⚠️ Could not load ruleset "Resources". Please check for an update from the developer.
I’m using the latest version of osu!. After installing the ruleset, I get the error shown in the title on both Linux (pop os!!) and Windows, and it says I should check the update from the developer.

Below is the log in windows .

How can I fix this issue?

[1764722167.runtime.log](https://github.com/user-attachments/files/23902091/1764722167.runtime.log)

bintis:
## Summary

- embed the official karaoke resource files directly in the ruleset assembly
- remove the separate `osu.Game.Rulesets.Karaoke.Resources` package and `DllResourceStore`
- stop copying the resource assembly into release output while preserving satellite resource assemblies
- add a regression test for resource lookup and assembly references

## Why

osu! scans assemblies named `osu.Game.Rulesets.*.dll` as ruleset plugins. The separately packaged `osu.Game.Rulesets.Karaoke.Resources.dll` is therefore discovered as a ruleset and produces the `Could not load ruleset "Resources"` error. Embedding the same resources into the actual ruleset assembly avoids that false plugin discovery.

## Testing

- `dotnet test osu.Game.Rulesets.Karaoke.Tests/osu.Game.Rulesets.Karaoke.Tests.csproj -c Release --filter FullyQualifiedName~KaraokeRulesetResourceTest` (1 passed)

Fixes #2337


andy840119:
"Could not load ruleset "Resources"" is a known issue for a long time.
I recommend find the way to combined dlls while compiling instead of just copy the resource file in here.
Because it might let the main repo become bigger faster if change the resource file frequently.

